PR #27, opened 2020-07-06
event data update - satyam bansal

satyam22:
Added my topic for July 12 event

### DIFF
--- a/events.html
+++ b/events.html
@@ -248,6 +248,34 @@
 					optimizing DOM manipulation and code
 				</li>
 			</ul>
+    </div>
+    <div class="session-info">
+			<h1 class="topic-name">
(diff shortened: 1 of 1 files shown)


events.html
```diff
@@ -248,6 +248,34 @@ <h3 class="speaker-info">Works at DocOn Technologies</h3>
 					optimizing DOM manipulation and code
 				</li>
 			</ul>
```
ankushdharkar, line 250: I think this should be removed, as you have closed it below

satyam22: it's GitHub file compare issue for closing "div" tag that's why it's showing on top.
if you check at end of code, it's not there.
I rechecked it to confirm.

PR #34, opened 2020-07-22
removed the swirls image from background

sumitd94:


takshch:
`pseudo element` is not removed. Only background property is removed. 
Issue-#39

PR #46, opened 2020-08-04
CSS enhancements to events page

tejasberi:


css/events.css
```diff
@@ -29,43 +29,42 @@
...
-  border-radius: 5px;
-  padding: 25px;
+  border-radius: 10px;
+  padding: 40px;
   display: block;
-  box-shadow: 0 20px 30px 0px rgba(228, 58, 99, 0.2);
   background: #fff;
+  -webkit-box-shadow: 2px 2px 18px 0px rgba(0, 0, 0, 0.75);
+  -moz-box-shadow: 2px 2px 18px 0px rgba(0, 0, 0, 0.75);
+  box-shadow: 2px 2px 18px 0px rgba(0, 0, 0, 0.75);
```
ankushdharkar, line 67: I don't think we need so many vendor prefixes @tejasberi. Please confirm the ones we should fly in

tejasberi: Since this is a plane HTML-CSS project, Please confirm if there is a auto-prefixer in the build/deploy cycle for this project. If not then its better to have those many vendor prefixes.

ankushdharkar: @tejasberi We don't need prefixes for box-shadow for most browsers nowadays: https://caniuse.com/#search=box-shadow

tejasberi: done

PR #55, opened 2020-10-02
Implemented Theme-changer toggle button

rahuldey613:
Added Dark mode Functionality
 Hope u will like it.

### DIFF
--- a/index.html
+++ b/index.html
@@ -57,6 +57,25 @@
         or our
         <a href="https://welcome.realdevsquad.com/faq"> FAQ section</a>.
       </h3>
+      <!-----Theme Changer--------->
+      
+      <ul>
+      <li>
+         <span>Dark</span>
+         <span>Light</span>
+      </li>
+    </ul>
+
+    <script src="https://code.jquery.com/jquery-3.3.1.js"></script>
+    <script type="text/javascript">
+        $(document).ready(function(){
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -57,6 +57,25 @@ <h3 class="info-note"> To know how this page is developed,
...
+
+    <script src="https://code.jquery.com/jquery-3.3.1.js"></script>
+    <script type="text/javascript">
+        $(document).ready(function(){
+          $('ul').click(function(){
+            $('ul').toggleClass('active')
+            $('section').toggleClass('dark')
+      })
+    })
+    </script>
```
ankushdharkar, line 77: @rahuldey613 Thanks for your PR. Theme changer is a great idea.

1. Use simple javascript instead of jQuery library
2. Just adding a class won't make the theme work

[1] We don't need to download extra jQuery for executing simple straightforward code. You should use simple javascript browser and DOM APIs to get this done. Please make the changes are re-request a review. Thanks!

[2] Your code only attaches a 'dark' class in the section. We would also need the accompanying CSS changes to make that work. I don't see that code in the PR or a mention of this in the description

ankushdharkar (changes requested):

swarajpure:
@rahuldey613 Are you going to make changes?

Issue #52, opened 2020-08-29
Show "Hello, <name>" instead of Login


ankushdharkar:
If the current user is already logged in, we want to show "Hello, <name>", instead of the the login button

<img width="600" alt="login" src="https://user-images.githubusercontent.com/1935403/91636522-f21b1100-ea1e-11ea-8666-ab4fc5296e80.png">

We can use the user details API to figure out the user's name. Please ask the backend team to update the [API contracts](https://github.com/Real-Dev-Squad/website-api-contracts)

swarajpure:
Issue solved. Refer #57 #62 and #63 for solution.
Closing this issue.

PR #57, opened 2020-10-05
Hello <first_name>!

swarajpure:
Fixes issue #52 

Hello, <first_name>!

Please review. Thank you!


index.html
```diff
@@ -13,11 +13,11 @@
 
 <body>
   <div class="content-wrapper">
-    <nav>
-      <a href='https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97' class="btn-login">
-        Login
-      </a>
-    </nav>
+    <div id="user-login">
```
ankushdharkar, line 16: Why change the `nav` into a `div`?
`nav` is a more semantic tag for html.

Read more here: https://css-tricks.com/why-how-and-when-to-use-semantic-html-and-aria/

index.js
```diff
@@ -0,0 +1,19 @@
+const userLogin = document.querySelector("#user-login");
+
+const setName = (name) => {
+    if (name) {
+        userLogin.innerHTML = `Hello, ${name}!`;    
+    }
+}
```
ankushdharkar, line 7: Is arrow function supported on all the browsers? Preferable to use ES5 syntax here?

index.js
```diff
@@ -0,0 +1,19 @@
...
+    fetch('https://staging-api.realdevsquad.com/users/self')
+    .then(res => res.json())
+    .then(res => {
+        setName(res.first_name)
+    })
+}
+
+fetchData();
+
+
```
ankushdharkar, line 19: Nit: Only one ending line please

index.js
```diff
@@ -0,0 +1,19 @@
...
+    if (name) {
+        userLogin.innerHTML = `Hello, ${name}!`;    
+    }
+}
+
+const fetchData = () => {
+    fetch('https://staging-api.realdevsquad.com/users/self')
+    .then(res => res.json())
+    .then(res => {
+        setName(res.first_name)
```
ankushdharkar, line 13: Nit: Maybe a more descriptive name next time? `setNameIfFound`

More read:
https://google.github.io/eng-practices/review/reviewer/standard.html
https://stackoverflow.com/questions/27810522/what-does-nit-mean-in-hacker-speak

ankushdharkar (changes requested):

index.html
swarajpure: Cool. Will keep this mind. The article is nice :)

index.js
swarajpure: IE and Opera mini don't support arrow functions. Fixed.

index.js
swarajpure: Done.

index.js
swarajpure: Done. Thank you!

index.js
```diff
@@ -0,0 +1,17 @@
+const userLogin = document.querySelector("#user-login");
```
ankushdharkar, line 1: Why leave the const in global space?

Why not:

```js
function setNameIfFound(name){
  const userLogin = document.querySelector("#user-login");
  if (name) {
    userLogin.innerHTML = `Hello, ${name}!`;    
  }
};
```

ankushdharkar (changes requested):
Small change please

index.js
```diff
@@ -0,0 +1,17 @@
...
+
+const fetchData = () => {
+    fetch('https://staging-api.realdevsquad.com/users/self')
+    .then(res => res.json())
+    .then(res => {
+        setNameIfFound(res.first_name)
+    });
+};
+
+fetchData();
```
ankushdharkar, line 16: While this works because you load your js later in the HTML, it may be worthwhile wrapping this in document loaded callback: https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event

ankushdharkar (changes requested):
What do you think about this?

index.js
swarajpure: Yeah, had not thought of that. Thanks!

index.js
swarajpure: Makes sense.

ankushdharkar (approved):
Thank you! Approved

index.js
ankushdharkar: Not fixed for `fetchData`

Issue #54, opened 2020-09-23
User profile information form


ankushdharkar:
When a new user comes to our website, and logins via Github, we redirect them to /goto which then decides where the user was before the login process began, and should take them back there. Currently, /goto always drops the user back to the www site

Upon login via Github for a new user, we create a temporary id for them, we want to change this into an RDS username with certain rules
- Starts with users's first name
- No special characters, except hyphens
- No funky unprofessional words in the username
- Avoid numbers

We want to create a page with a form where a new user can select a username, and fill out or correct their first name, last name, YoE, Company, etc (We can pull in the initial name from Github during login)

We need to create this page to allow new users to sign up to our platform

sujitmohanty:
1. Regex mis-validates username
2. Database setup possibly incomplete.
3. Failure at acquiring desired output when making api calls despite cookie being present.

PR #64, opened 2020-10-14
User profile information form #54 (WIP)

sujitmohanty:
Changes-
1. User form with input field validation
2. Features function to modify logged in user data
3. Button attribute set to disabled & opacity to 0.5 respectively if any mandatory input fields are empty.
4. Incorporated correct Regex logic but code implementation doesn't yield desired output.


user-form.html
```diff
@@ -0,0 +1,114 @@
...
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,400;0,500;0,700;1,400&display=swap"
+      rel="stylesheet"
+    />
+    <link
+      rel="stylesheet"
+      href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.14.0/css/all.min.css"
+    />
+  </head>
+  <body>
+    <nav>
```
swarajpure, line 18: Please check `index.html` for new code for navbar. That PR would have been merged after you had started working on it.

swarajpure (changes requested):
Amazing job in there! So much to learn from your code!
Suggesting some changes. Please check. Thank you!! 😄

js/user-form.js
```diff
@@ -0,0 +1,109 @@
...
+    linkedin_id,
+  ]);
+
+  updateUserData();
+});
+
+const id = username.value;
+const updateUserData = async () => {
+  try {
+    let res = await fetch(`http://localhost:3000/users/${id}`, {
```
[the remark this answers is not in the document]
sujitmohanty: oh no, I am sorry @swarajpure ! It should have been this URL that you just mentioned.

js/user-form.js
```diff
@@ -0,0 +1,109 @@
...
+  formGroup.className = 'form-group success';
+}
+
+// Username Validator
+function checkUsername(input) {
+  //Check presence of first_name
+  if (!username.value.trim().startsWith(first_name.value.toLowerCase())) {
+    showError(
+      username,
+      'Username should start with your first name. Please avoid using funky words as suffix'
```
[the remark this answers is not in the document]
sujitmohanty: No, @swarajpure ! I believe the username term would be coined starting from the respective user's firstName and then optionally a hyphen or alphabetical suffixes.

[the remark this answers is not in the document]
swarajpure: What I see from the code is that the user  is not allowed to use a username that does NOT start with their first name. So the error should show that the user should not have a username which doesn't START with their first name and hence I was asking if you meant to use `prefix` here instead of `suffix`.

user-form.html
```diff
@@ -50,7 +50,12 @@ <h2>Account Details</h2>
 
           <div class="form-group">
             <label for="yoe">Experience (in Years): </label>
-            <input type="text" id="yoe" name="yoe" />
+            <input
+              type="number"
+              id="yoe"
+              name="yoe"
+              placeholder="'0' if student"
```
swarajpure, line 57: Thank you for considering students 😁
Can we have similar for company name?

swarajpure (changes requested):
Please solve merge conflict. Suggesting minor changes.
Also, the API call will fail. We need to discuss about it. One approach can be to call `/users/self` API to get the `id` and then pass it to `PATCH` request.

sujitmohanty:
Features Added:
1. Input boxes now have focus
2. Placeholder text is present in all the input boxes. 
3. Added hint text to username.
4. Email and Phone Number input boxes are added with hint text (disclaimer) stating their details would be private & are being taken to notify them of future events.

js/user-form.js
```diff
@@ -0,0 +1,134 @@
...
+  } catch (error) {
+    console.error(error);
+  }
+};
+
+const updateUserData = async () => {
+  try {
+    let res = await fetch(`https://staging-api.realdevsquad.com/users/${id}`, {
+      method: 'PATCH',
+      credentials: 'include',
```
swarajpure, line 127: 1. What are we sending with the PATCH request? The data to be updated in the databse needs to be sent in the body of the PATCH request. Nothing is being sent in this request to be updated.
2. The updateUserData and getUser functions are being fired simultaneously/asynchronously, that's why the `id` sent in the URL of PATCH req will be `undefined` in this case. We need to have returned `id` from the getUser function before we actually do the PATCH req. The flow of the requests needs to be proper otherwise the form won't work as expected. Please test on your local machine before you push it so ensure that the functionality is proper.

I hope I'm making sense 😅
Please let me know if I'm confusing you.

js/user-form.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+    const data = await res.json();
+    console.log(data);
```
swarajpure, line 130: Are you sure about keeping the console logs in the final code?

swarajpure (changes requested):
The flow of API calls needs to be properly set otherwise the form will not function as expected. Please see to it.

PR #67, opened 2020-11-02
Added a showcase area for new updates and notifications

ankushdharkar:
This area will be useful to show some of the new things happening across our sites

image

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -61,12 +61,12 @@ nav {
   padding: 2px 8px;
   font-weight: bold;
   font-size: 0.8em;
-  margin-right: 5px;
+  margin: 0 5px;
   border-radius: 4px;
 }
 
 .interesting-notifications > .notif a {
-  margin:0 10px;
+  margin: 0 5px 0 10px;
   font-size: 0.85em;
 }
 
--- a/css/main.css
+++ b/css/main.css
@@ -41,16 +41,45 @@ nav {
   color: #041484;
 }
 
+.interesting-notifications {
+  padding: 10px 8px;
+  border: 1px solid #bdbdbd;
+  border-radius: 10px;
+  background-color: #fafafa;
+  margin: 20px 0 50px;
+}
+
+.interesting-notifications > .notif {
+  font-size: 16px;
+  line-height: 1.5;
+  margin: 0;
+}
+
+.interesting-notifications > .notif span {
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -19,9 +19,17 @@
         </a>
     </nav> 
 
+    <section class="interesting-notifications">
+      <p class="notif">
+        <span>New</span>
+        <strong>Crypto</strong>
+        -- New updates to the user dashboard
+        <a href="https://crypto.realdevsquad.com"> Check it out! ➡️ </a>
```
swarajpure, line 27: What if we take the arrow emoji out of anchor tag?
The underline for anchor tag is going to the right of arrow emoji and it is looking a little weird. I initially thought it's a dot.
image

swarajpure (commented):
Thanks for this feature! 😄 
Have some minor doubts. Please check.

css/main.css
```diff
@@ -41,16 +41,45 @@ nav {
...
+}
+
+.interesting-notifications > .notif span {
+  background: linear-gradient(90deg,#e30162 25%,#ed579c 90%);
+  color: white;
+  padding: 2px 8px;
+  font-weight: bold;
+  font-size: 0.8em;
+  margin-right: 5px;
+  border-radius: 4px;
```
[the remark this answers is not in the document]
ankushdharkar: Makes sense. Thanks @swarajpure

index.html
ankushdharkar: This is how my emoji looks: 
image

ankushdharkar: I want the emoji to be clickable

swarajpure: Even for emoji in your case, there's a little dot at the bottom right of the emoji. It's looking a bit weird to me. Plus, most of the users will come from Windows, it is looking a bit weird on Windows. Can we add a different id to it, still inside anchor tag but removing it's underline from CSS?

Just my opinion. Maybe we could compare the two - emoji not underlined and emoji underlined and see which one looks better?

ankushdharkar: I changed the underline to none, and using a color transition now to denote that the link is clickable

ankushdharkar:
image

index.html
swarajpure: Cool. Thank you!

swarajpure (approved):
All looks good to me now. I had suggested some changes which were fixed.

PR #70, opened 2020-11-07
Adding navbar on home-page and events-page

akankshadharkar:
Screenshot 2020-11-07 at 19 26 25

**Next phase** - Responsiveness for mobile screens

### DIFF
--- a/events.html
+++ b/events.html
@@ -32,7 +32,8 @@
       Web Mini-Conference | Important concepts a Web Developer should know
     </title>
     <link rel="stylesheet" href="./css/main.css" />
-    <link rel="stylesheet" href="./css/events.css" />
+		<link rel="stylesheet" href="./css/events.css" />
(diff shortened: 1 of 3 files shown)


events.html
```diff
@@ -32,14 +32,26 @@
       Web Mini-Conference | Important concepts a Web Developer should know
     </title>
     <link rel="stylesheet" href="./css/main.css" />
-    <link rel="stylesheet" href="./css/events.css" />
+		<link rel="stylesheet" href="./css/events.css" />
+		<link rel="stylesheet" href="./css/navbar.css">
```
ankushdharkar, line 36: Why the extra indentation?

ankushdharkar (approved):

events.html
akankshadharkar: There isn't but github shows it that way 
Ah, must be that my tab is set to 4 instead of 2 spaces 🤦

PR #71, opened 2020-11-08
Implementing user form to signup a profile

### LINKED ISSUE #54: User profile information form

ankushdharkar:
This allows the user to create profile for the very first time

Closes #54


js/user-form.js
```diff
@@ -0,0 +1,155 @@
...
+    DESIGNATION,
+    GITHUB_ID,
+    LINKEDIN_ID,
+    TWITTER_ID
+  ]);
+
+  console.log(dataToSend);
+
+  try {
+    const res = await fetch(`AWAITING_API__https://staging-api.realdevsquad.com/users/self`, {
```
ankushdharkar, line 143: @Rucha1499 I am ready to start using the API. Let me know when it's ready please

js/user-form.js
```diff
@@ -0,0 +1,155 @@
...
+    TWITTER_ID
+  ]);
+
+  updateUserData();
+});
+
+function getDataToSend (arrKeys) {
+  return arrKeys.reduce(function(acc, k) {
+    const el = document.getElementById(k);
+
```
sumitd94, line 116: I think we can be more specific with the arguments names being passed to the anonymous function on the `reduce`
Like if I directly jump on this function, I have no clue what `acc` and `k` means

js/user-form.js
```diff
@@ -0,0 +1,155 @@
...
+    EMAIL,
+    PHONE, 
+    YOE,
+    COMPANY,
+    DESIGNATION,
+    GITHUB_ID,
+    LINKEDIN_ID,
+    TWITTER_ID
+  ]);
+
```
sumitd94, line 139: Lets create a constant at the top and store this argument which is being passed to `getDataToSend` in it and then use that constant as argument.
similarly for the code on line no 98

js/user-form.js
```diff
@@ -0,0 +1,155 @@
...
+  console.log(dataToSend);
```
swarajpure, line 140: Are we keeping this console log statement?

ankushdharkar: No. I was just waiting for the API to start working. Will remove this in the final version. Thanks!

js/user-form.js
ankushdharkar: Great point @sumitd94 

Array.prototype.reduce has a pretty standard sequence according to the positions.
acc - accumulator
k - key

sumitd94: Yep

PR #73, opened 2020-11-09
Disabled user profile image in input

### LINKED ISSUE #72: User profile picture from website-static
Until we have the S3 bucket working, we will use `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members/<username>/img.png`

Repo here: https://github.com/Real-Dev-Squad/website-static/

ankushdharkar:
The user profile picture will come from the website-static repo. We will use the username as the key for selecting the profile image

Closes #72

### DIFF
--- a/user-form.html
+++ b/user-form.html
@@ -170,7 +170,8 @@
               type="text"
               id="profile_pic"
               name="profile_pic"
-              placeholder="Link to your profile picture"
+              placeholder="Your profile picture comes from website-static"
+              disabled
             />
             <small>Error Message</small>
           </div>
--- a/css/user-form.css
+++ b/css/user-form.css
@@ -55,6 +55,10 @@ body {
   border-color: #777;
 }
 
+.form-wrap .form-group input:disabled {
+  background: #B0BEC5;
+}
+
 .form-group.success input {
   border-color: var(--success-color);
   border-width: 2px;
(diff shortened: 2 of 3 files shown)


user-form.html
```diff
@@ -170,7 +170,8 @@ <h2>Account Details</h2>
               type="text"
               id="profile_pic"
               name="profile_pic"
-              placeholder="Link to your profile picture"
+              placeholder="Your profile picture comes from website-static"
```
swarajpure, line 173: What if we could link the `website-static` repo? (For someone who doesn't know what is `website-static`)

ankushdharkar: Good idea, let's do that as a follow-up feature

This entire www page is quite temporary. We will move it to an Ember app in future

swarajpure: I'm sure if its possible or no since it's placeholder (plain text I suppose)

ankushdharkar: Placeholder is just dead text, but we can always provide a link under the input box. I like your idea.

swarajpure: Cool ⚡

swarajpure (approved):

user-form.html
ankushdharkar: @swarajpure https://github.com/Real-Dev-Squad/website-www/issues/76

PR #80, opened 2020-11-25
Convert "login" link to "Sign in with Github" button

Khanjan04:
Convert "login" link to "Sign in with Github" button.


css/main.css
```diff
@@ -32,6 +32,19 @@ nav .btn-login {
   float: right;
 }
 
+nav .button{
+  font-size: 20px;
```
swarajpure, line 36: Doesn't the font size look too huge compared to rest of the page? Please try reducing the size and see which one looks better. After changing the font-size you'd also have to change the padding to compensate.
image

css/main.css
```diff
@@ -32,6 +32,19 @@ nav .btn-login {
...
+  border-radius: 06px;
```
swarajpure, line 37: `6px` would be more readable than `06px`

index.html
```diff
@@ -36,7 +36,7 @@
           class="btn-login"
           href='https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97'
         >
-          Login
+        <button class="button">Sign In With Github <img class="icon" src="img/github.png" alt="Github Icon" /></button>
```
swarajpure, line 39: Wouldn't something like `button class="login-btn-txt"` make more sense?
The class name should be self explanatory about it's functioning, right?

Same for img as well. `img class="logo"` would be more self explanatory.

swarajpure (changes requested):
Thank you for this pull request! 🎉 
Requesting some minor changes. Please see to it.

css/main.css
Khanjan04: Okay let me do that

css/main.css
Khanjan04: ok

index.html
Khanjan04: Ok. Let me change that

Khanjan04: Have done the changes. Please check them.

swarajpure (approved):
Great job! Thank you! 🎉 😄

PR #86, opened 2021-01-24
Redirect to signup form if incompleteUserDetails

### LINKED ISSUE #65: Flow for user profile information form
Currently the form has no connection with the website, the user will manually have to go `/user-form` to be able to view and fill the form.

The flow should be like:
1. Users click on login, they complete the login procedure from GitHub, get redirected back to our website.
2. They now has the cookie set in their browser for `realdevsquad.com`. They are logged in, but have not filled in all the required fields, i.e. they have incomplete profile.
3. We are already doing a call to `/users/self` API to get the users' first name for `Hello, <first_name>` (please refer `/index.html`), but if the `isIncompleteUser` is true, their first name wouldn't be shown, **now** they should be redirected to this form.
4. Since we are already calling the `self` API, we can access to the users' `id` field. We can pass that to the `PATCH /users/:id` API and get the details saved to the database.

ankushdharkar:
Closes #65


prakashchoudhary07 (approved):

swarajpure (dismissed):
Cool! Thank you! 🎉

prakashchoudhary07 (approved):

swarajpure (approved):

PR #93, opened 2021-03-06
Ref(index.html): redirection added to signup form

SagarBajpai:
Post Github signup user will be redirected to the Signup form to fill in further details.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -23,6 +23,9 @@ function fetchData () {
     })
     .then(res => res.json())
     .then(res => {
+        if(res.incompleteUserDetails){
+            return window.location.replace('https://my.realdevsquad.com/signup')
+        }
         setUserGreeting(res.username, res.first_name);
     });
 };


js/main.js
```diff
@@ -23,6 +23,9 @@ function fetchData () {
     })
     .then(res => res.json())
     .then(res => {
+        if(res.incompleteUserDetails){
+            return window.location.replace('https://my.realdevsquad.com/signup')
+        }
```
swarajpure, line 28: Wouldn't it be better to alert the user where the user is now being taken to? We could show an alert saying "You'll now be redirected to a form where you can complete your profile"
What do you think? 🤔

SagarBajpai: Ya we can. The only thing is alert will stop the redirection and ask for the user's response to make the redirection happen. The user may bounce back and probably will not do the signup. 
The way right now, user will simply redirect to signup post the github auth. S M O O T H ................

swarajpure: This is good per discussion with @ankushdharkar

swarajpure (approved):

Issue #102, opened 2021-03-22
Set up pre-commit hook for checking formatting


swarajpure:
We need pre-commit hook (which runs before the `git commit` command) to check for formatting problems in the files. This is to be done to ensure we are following consistent formatting throughout the codebase.

We'll be using this package: https://www.npmjs.com/package/pre-commit
Fairly simple to implement. You can read up the documentation for details related to installation and using it.

Script to be added in pre-commit hook: `check`

If you want to work on this issue, please comment down. It can be then assigned to you 🙂

rajakvk:
I like to take this up. Please assign. @swarajpure

swarajpure:
#104

PR #103, opened 2021-03-22
[feature/hook-format] Set up pre-commit hook for checking formatting

rajakvk:
Added pre-commit hook for formatting using "prettier".


swarajpure (changes requested):
We do not require so many packages. Just one package called `pre-commit` does the job
Please read the issue description #102

swarajpure:
Merged #104

PR #104, opened 2021-03-23
feature/hook-format using bare minimum required

rajakvk:
Added only required


swarajpure (approved):
Looks good!
Thanks for this one! 😁

PR #128, opened 2021-06-24
Setting up prettier to enable semicolons, force quotes as single quotes, tabWidth as 2 and to enable trailing commas

anshika29:


swarajpure (approved):
Thanks for the quick one!

Congratulations on your first pull request in Real Dev Squad! 🥳

PR #129, opened 2021-06-25
Added Check Script

himanshukm2301:


.github/workflows/percy.yml
```diff
@@ -33,3 +33,20 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install
+    - run: npm run build --if-present
+    - run: npm test
```
swarajpure, line 52: We just need `npm install` and `npm run check`
We don't need `build` and `test`

himanshukm2301: Updated

swarajpure (changes requested):
Great! Thanks 🎉 

Just small changes:
image

These don't make much sense, right? Can you see where they're coming from and how to fix them? First one is okay, second one should say something like "Checking formatting"

himanshukm2301:
Sure, I'll look into it!

swarajpure (approved):
Looks good to me!
Thank you! 🎉

Issue #135, opened 2021-07-13
Fix the member animation 


Vividh25:
**Bug**: The member animation in the page suddenly jumps back to the starting state when a particular part of the screen is scrolled and doesn't feel very smooth. Look at the attached video for more clarification. 

https://user-images.githubusercontent.com/61474917/125414159-d54be247-ae84-4770-93ab-4e791b521332.mp4

rahilkapoor:
Assign this to me, will try to fix it.

PR #136, opened 2021-07-19
Fixed Animation (Issue: Fix the member animation #135)

rahilkapoor:
Fixed the Bug in the Animation Issue #135.


js/animate.js
```diff
@@ -41,19 +36,19 @@ const removeClassName = () => {
 };
 
 memberList.forEach((member) => {
-  member.style.setProperty(
-    '--s',
-    `${randomNumberWithinRange(0.1, 0.6) * responsiveConstant + 0.9}`,
-  );
+  member.style.setProperty('--s', `${randomNumberWithinRange(0.9, 1.8)}`);
```
harshith-venkatesh, line 39: good update :)

js/animate.js
```diff
@@ -41,19 +36,19 @@ const removeClassName = () => {
...
+    if (element.isIntersecting) {
+      addClassName();
+    } else {
+      removeClassName();
+    }
+  });
+};
+
+const userViewport = document.querySelector('#members');
+observer = new IntersectionObserver(handleIntersection);
```
harshith-venkatesh, line 53: Why are we not declaring observer here?
could we go with const since it is an object and we are not mutating it

harshith-venkatesh (commented):
Great work @rahilkapoor , I got to understand more on Intersection Observer API with this practical example

js/animate.js
rahilkapoor: Thanks😅

js/animate.js
```diff
@@ -41,19 +36,19 @@ const removeClassName = () => {
...
-  let show = false;
-  if (!show && window.scrollY > height * 0.8 && window.scrollY < height * 1.8) {
-    show = true;
-    addClassName();
-  } else {
-    removeClassName();
-    show = false;
-  }
-});
+let handleIntersection = (entries) => {
```
harshith-venkatesh, line 42: Function definition - const should go here

js/animate.js
```diff
@@ -41,19 +36,19 @@ const removeClassName = () => {
...
+  entries.forEach((element) => {
+    if (element.isIntersecting) {
+      addClassName();
+    } else {
+      removeClassName();
+    }
```
sumitd94, line 48: Since we are only using the `isIntersecting` property from the `element` object, let's destructure it and use it. Also let's use the shorthands for if-else

```javascript
entries.forEach(({ isIntersecting }) => {
 if (isIntersecting) addClassName();
 else removeClassName();
});
```

Now why I am telling you to do this?

- The code now looks concise and easier to read and understand.

sumitd94 (changes requested):
One small change, please

js/animate.js
rahilkapoor: Yeah, I see...I knew this but forgot to implement it. Thanks for the reminder!👍

harshith-venkatesh: Good catch @sumitd94 , Even I will remember and implement in my future PRs , thanks

harshith-venkatesh (approved):

harshith-venkatesh:
Good job, approved from my end

rahilkapoor:
> Good job, approved from my end

Thanks ✌✔

sumitd94 (approved):
Thanks, Looks good now!

Issue #138, opened 2021-07-25
Make the member images on the page appear dynamically.


Vividh25:
Problem Description: 
Currently there are 5 image tags hard-coded in the `index.html` page (see  line no. 170 in [this](https://github.com/Real-Dev-Squad/website-www/blob/develop/index.html)). We need to add these 5 images dynamically into the DOM instead of hard-coding them.

rahilkapoor:
I would like to take this

PR #141, opened 2021-08-07
Make the member images on the page appear dynamically. #138

rahilkapoor:
Fixed for Mobile Version

mob


index.html
```diff
@@ -6,9 +6,9 @@
     <link rel="shortcut icon" href="./img/favicon.ico" type="image/x-icon" />
     <link rel="icon" href="./img/favicon.ico" type="image/x-icon" />
     <title>Real Dev Squad</title>
-    <link rel="stylesheet" href="./css/main.css" />
-    <link rel="stylesheet" href="./css/animate.css" />
-    <link rel="stylesheet" href="./css/navbar.css" />
+    <link rel="stylesheet" href="./css/main.css" defer />
```
harshith-venkatesh, line 9: defer is used for scripts https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

index.html
```diff
@@ -6,9 +6,9 @@
...
+    <link rel="stylesheet" href="./css/animate.css" defer />
```
harshith-venkatesh, line 10: https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ

harshith-venkatesh (changes requested):

index.html
```diff
@@ -319,13 +319,7 @@ <h2 class="title-bullet-points">Soft Skills</h2>
...
         </section>
 
-        <section class="member-imgs" id="members">
-          <img class="member-img member_animation" alt="member-img" />
-          <img class="member-img member_animation" alt="member-img" />
-          <img class="member-img member_animation" alt="member-img" />
-          <img class="member-img member_animation" alt="member-img" />
-          <img class="member-img member_animation" alt="member-img" />
-        </section>
+        <section class="member-imgs" id="members"></section>
```
swarajpure, line 322: Empty tags can be self closing
`<section class=.......... />`

js/main.js
```diff
@@ -28,8 +28,17 @@ const selectRandom = (memberImgArr, n) => {
   return result;
 };
 
+const numOfMembers = 5;
+let memberSection = document.getElementById('members');
+for (let i = 0; i < numOfMembers; i++) {
+  const memberTags = document.createElement('img');
+  memberTags.classList.add('member-img', 'member_animation');
+  memberTags.setAttribute('alt', 'member-img');
```
swarajpure, line 36: `Alt` text could non-technical? It should be simple english, if the images fail to load, hovering upon them we should see that text right?

swarajpure (changes requested):
Thanks!
Requesting some minor changes, please check!

index.html
rahilkapoor: We are dynamically adding tags later...So we need the closing tag here!

js/animate.js
```diff
@@ -8,17 +8,28 @@ const randomNumberWithinRange = (min, max) => {
...
+  const deg = n * ((360 / numOfMembers) * (Math.PI / 180));
   return 160 * Math.sin(deg);
 };
 
 const memberList = document.querySelectorAll('.member_animation');
 
+const handleIntersection = (entries) => {
+  entries.forEach(({ isIntersecting }) => {
+    if (isIntersecting) addClassName();
+    else removeClassName();
```
swarajpure, line 25: How about?
```JS
isIntersecting ? addClassName() : removeClassName()
```

rahilkapoor: This was a really good one! Thanks, Swaraj!

swarajpure (approved):
Looks good to me! 🔥

Approving from my side

Vividh25 (approved):
Great work here!

Issue #116, opened 2021-06-07
Create art creator feature


ankushdharkar:
Whatever you type on the left hand side box shows up on the right side.
This task already has two PRs associated with it. Check [here](https://github.com/Real-Dev-Squad/website-www/pull/145) and [here](https://github.com/Real-Dev-Squad/website-www/pull/143).

[Figma Link](https://www.figma.com/file/bbT6ouwsrjYzs6G4XDSn3a/CSS-Art?node-id=0%3A1)

ToBuild-2021:
commenting to get the task assigned.

ankushdharkar:
@ToBuild-2021 I have assigned you this existing issue: https://github.com/Real-Dev-Squad/website-dashboard/issues/33

You should join our discord server to stay more in touch (it's free): https://discord.gg/sXazaRz6

ToBuild-2021:
yeah 
Working on that .

PR #145, opened 2021-08-27
Implemented art feature

PayalHake1402:
issue : Create art creator feature #116

Updated code.


rohan09-raj:
There were no further developments That's why this has been already taken care of and merged with PR #155

PR #157, opened 2021-10-03
added max-width to textarea tag

abhinav1590:


harshith-venkatesh (commented):
Are all the issues addressed?

abhinav1590:
> Are all the issues addressed?

Yes

js/artFeature.js
```diff
@@ -15,5 +15,5 @@ const creatingArtFromHtml = () => {
 };
 
 document
-  .querySelector('html-code')
```
harshith-venkatesh, line 18: Appreciate your patience.

Could you please see how can we utilize `querySelector` itself instead of changing into `getElementById`
https://dev.to/eidorianavi/queryselector-vs-getelementbyid-gm1
One of the available resource
I would still recommend to go through MDN docs though

harshith-venkatesh (changes requested):

rohan09-raj:
There was no further work and comments were not addressed from very long, that's why closing this PR

PR #160, opened 2021-10-14
Dark-mode feature for Main and Events site

### LINKED ISSUE #154: Dark Mode for www-site
Implementing Dark Mode for www-site
Working Member - Rohan

rohan09-raj:
Closes #154 

https://user-images.githubusercontent.com/78433013/137275392-22fb4e15-a782-462d-9bf9-58d1610d27bc.mp4


css/main.css
```diff
@@ -1,11 +1,35 @@
...
+  --color-text: #fff;
+  --color-text-secondary: #000;
+  --color-what-is-title-bg: #1d1283;
+  --color-why-description-text: #3d3d3d;
+  --color-social-media-links-bg: #fff;
+  --color-links: #1d1283;
+  --color-events: #e30162;
+}
+
+.dark-theme {
```
akshay1502, line 12: Could we have a separate css file, where we will have `lightTheme` and `darkTheme` classes? And let's import those classes wherever needed.

css/main.css
```diff
@@ -1,11 +1,35 @@
+:root {
```
akshay1502, line 1: creating a separate class called `lightTheme` would be better here than directly doing in :root . Like you did for `darkTheme`.

css/main.css
```diff
@@ -177,12 +205,12 @@ nav .user-greet .user-greet-msg {
 }
 
 .social-media-links > span {
-  color: #e20665;
+  color: #e30162;
```
akshay1502, line 208: You have added #e30162 as `--color-events`, so could we use that here?

js/darkmode.js
```diff
@@ -0,0 +1,45 @@
+onload = onload();
+const icon = document.getElementById('icon');
+icon.onclick = function () {
+  const element = document.body;
+  element.classList.toggle('dark-theme');
+  if (element.classList.contains('dark-theme')) {
+    icon.src = 'img/sun.png';
+    setCookie('theme', 'dark', 30);
```
akshay1502, line 8: we have decided to use `theme` as name of cookie, so could we just directly mention it In document.cookie and not over here.

js/darkmode.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function setCookie(name, value, days) {
+  var expires = '';
+  if (days) {
+    var date = new Date();
+    date.setTime(date.getTime() + days * 24 * 60 * 60 * 1000);
+    expires = '; expires=' + date.toUTCString();
+  }
+  document.cookie = name + '=' + (value || '') + expires + '; path=/';
```
akshay1502, line 33: add the domain too here `domain=.realdevsquad.com`

akshay1502: let's make use of template strings here for better readability and use.

css/main.css
rohan09-raj: yes we can use that here.

js/darkmode.js
rohan09-raj: I guess we should make the function for setting the cookie reusable so that if we further need we can directly use it, @shubham-y is also doing the same.

akshay1502: Great. I will do the same in my PR.

js/darkmode.js
rohan09-raj: Can you please give a example here ??

shubham-y: @rohan09-raj Please refer the following example:
``` document.cookie = `${name}=${value}; expires=${expires}; domain=${domain}; path=/`;```
Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

Comment (rohan09-raj) [reply js/darkmode.js:33]: > @rohan09-raj Please refer the following example: `` document.cookie = `${name}=${value}; expires=${expires}; domain=${domain}; path=/`; `` Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

Thank you

File: css/main.css (hunk 2 above)
Comment (rohan09-raj) [reply css/main.css:1]: We can do that but when website initially loads (i.e. when it is not having a cookie) there will be no theme which will be applied to pages and it will be all black and white and if we do this by adding a class (like add a class when the site loads), so if someone has slow internet speed (we know that javascript loads in the end), so there will be delay in apply the theme for few miliiseconds which will lead to bad UX.

File: js/darkmode.js [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+    setCookie('theme', 'light', 30);
+  }
+};
+
+function onload() {
+  const themeIcon = document.getElementById('icon');
+  if (getCookie('theme') === 'light') {
+    document.body.classList.remove('dark-theme');
+    themeIcon.src = 'img/moon.png';
+  } else if (getCookie('theme') === 'dark') {
```
akshay1502, line 20: No need of `else if` here. We can directly go with else.

rohan09-raj: Just wanted to be more specific for that cookie name.

shubham-y: @rohan09-raj Store the value returned from the `getCookie` function in a variable and use it for comparision rather than calling it twice.

PR #171, opened 2021-11-02
centered as per login button

iRahulP:
Centered as per login button, and text visuals changed to white.

Will try to update the final one once fixed for the mobile version.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -52,9 +52,8 @@ nav .logo {
 }
 
 nav .user-greet {
-  display: none;
   font-size: 20px;
-  margin-top: 20px 30px 0 0;
+  margin: 20px 30px 0 0;
   float: right;
 }
 
--- a/css/main.css
+++ b/css/main.css
@@ -53,7 +53,7 @@ nav .logo {
 
 nav .user-greet {
   font-size: 20px;
-  margin: 20px 30px 0 0;
+  margin: 20px 30px 0 0 !important;
   float: right;
 }
 
--- a/css/main.css
+++ b/css/main.css
@@ -54,7 +54,8 @@ nav .logo {
 nav .user-greet {
   display: none;
   font-size: 20px;
-  margin-top: 10px;
+  margin-top: 20px 30px 0 0;
+  float: right;
 }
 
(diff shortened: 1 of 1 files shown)


css/main.css
```diff
@@ -54,7 +54,8 @@ nav .logo {
 nav .user-greet {
   display: none;
   font-size: 20px;
-  margin-top: 10px;
+  margin-top: 20px 30px 0 0;
```
harshith-venkatesh, line 57: Could you please look into this...
It should be `margin`

https://developer.mozilla.org/en-US/docs/Web/CSS/margin

css/main.css
```diff
@@ -54,7 +54,8 @@ nav .logo {
 nav .user-greet {
   display: none;
```
harshith-venkatesh, line 55: Why do we have `display:none` here?

harshith-venkatesh (changes requested):

css/main.css
iRahulP: It was already there initially from the initial author for the code base, Should I update or just remove it.

iRahulP:
snap

Current preview after update, also removed the display:none as it was already overridden by the root property.

Vividh25:
> 
> 
> snap
> 
> Current preview after update, also removed the display:none as it was already overridden by the root property.

Can you post a screenshot for mobile version ?

iRahulP:
mobile

Current layout for mobile view.

Vividh25:
> 
> 
> mobile
> 
> Current layout for mobile v

css/main.css
```diff
@@ -52,9 +52,9 @@ nav .logo {
 }
 
 nav .user-greet {
-  display: none;
   font-size: 20px;
-  margin-top: 10px;
+  margin: 20px 30px 0 0 !important;
```
harshith-venkatesh, line 56: Sorry for the late response, could we look into the way , where we don't use `!important` please

harshith-venkatesh (changes requested):

rohan09-raj:
This has been taken care of in PR #178 as there were new design changes in mobile view.

PR #186, opened 2021-12-23
fix: navigation to user profile page

rajat-mehra05:
List of the changes done-
- Fixed navigation route to user profile page i.e. `https://my.realdevsquad.com/profile` inside `index.html` after user clicks on
   `Hello, user-name`.
- replaced `realdevsquad.com` with `www.realdevsquad.com` in `events.html`. 
- fixed link `https://crypto.realdevsquad.com/auction` inside `releases.html.

Let me know if I need to make further changes.

### DIFF
--- a/css/navbar.css
+++ b/css/navbar.css
@@ -65,6 +65,10 @@ nav li a:hover {
   margin: 25px;
 }
 
+.user-greet a {
+  padding: 0;
+}
+
 .user-profile-pic {
   width: 32px;
   height: 32px;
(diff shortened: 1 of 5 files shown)


rajat-mehra05:
There are some issues in my PR. I will create a new PR for the same and add updated changes.

rohan09-gupta:
> There are some issues in my PR. I will create a new PR for the same and add updated changes.

@rajat-mehra05 This conflicts are normal to have. You should know how to solve them as this will be usual work at your workplace.

PR #188, opened 2021-12-24
fix: navigation to user profile page

### LINKED ISSUE #162: Navigate to `my.realdevsquad.com`
If someone clicks on `Hello, <username> and profile picture`, the user should be taken to https://my.realdevsquad.com/profile
image

rajat-mehra05:
Closes #162 
1. Added link to redirect user to `https://my.realdevsquad.com/profile` page on clicking `Hello, user-name`
2. Replaced `realdevsquad.com` with `www.realdevsquad.com` in `events.html` and `releases.html` files
3. Fixed `https://crypto.realdevsquad.com/auction` link

### DIFF
--- a/css/navbar.css
+++ b/css/navbar.css
@@ -50,7 +50,7 @@ nav li a:hover {
   display: none;
 }
 
-.btn-login-text { 
+.btn-login-text {
   border-radius: 6px;
   padding: 5px;
   background-color: #1d1283;
@@ -68,11 +68,7 @@ nav li a:hover {
 .user-greet a {
   text-decoration: none;
   color: #fff;
-}
-
-.user-greet a:hover {
-  color: #49a82e;
-  cursor: pointer;
+  padding: 0;
 }
 
 .user-profile-pic {
--- a/css/navbar.css
+++ b/css/navbar.css
@@ -62,7 +62,7 @@ nav li a:hover {
 .user-greet {
   display: none;
   color: #ffffff;
-  margin: 25px;
+  margin: 23px;
 }
 
 .user-greet a {
--- a/css/navbar.css
+++ b/css/navbar.css
@@ -50,7 +50,7 @@ nav li a:hover {
   display: none;
 }
 
-.btn-login-text {
+.btn-login-text { 
   border-radius: 6px;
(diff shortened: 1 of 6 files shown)


index.html
```diff
@@ -61,10 +61,15 @@
...
-          <div class="user-greet-msg">Hello, User!</div>
-          <img class="user-profile-pic" />
-        </div>
+
+        
+        <div class="user-greet">        
+            <a href="https://my.realdevsquad.com/profile">
+            <div class="user-greet-msg">Hello, User!</div>
+            <img class="user-profile-pic" />      
+          </a>
```
rohan09-raj, line 70: Please fix the formatting here.
```
<a href="https://my.realdevsquad.com/profile">
  <div class="user-greet-msg">Hello, User!</div>
  <img class="user-profile-pic" />      
</a>
```
It should be like this.

css/navbar.css
```diff
@@ -65,6 +65,16 @@ nav li a:hover {
   margin: 25px;
 }
 
+.user-greet a {
+  text-decoration: none;
+  color: #fff;
+}
+
+.user-greet a:hover {
+  color: #49a82e;
```
rohan09-raj, line 74: Why are we changing the color on hover ??

rohan09-raj (changes requested):
@rajat-mehra05 
1- Please check there are two login buttons in the `index.html` so the thing is we are using two buttons for login, one for mobile view and one for desktop view. That's why you were not able to click on the user greet when the screen size was greater than `970px`.
2- Also, please add the anchor tags to user-greet in `releases.html` and `events.html`, as these pages are also part of this site and they to have to navbar similar to `index.html`

css/navbar.css
rajat-mehra05: The other buttons on navbar seem to have the same property, that's why I had put `hover` property there.

rohan09-raj: No for this button we don't need any color change, only cursor changing to `pointer` will do the job.

rohan09-raj (approved):
Looks great to me. Approving the changes, thanks for sticking around.

PR #200, opened 2022-01-06
Fixed layout shift issue on reload

### LINKED ISSUE #198: Fix the content shifting while Page gets Loaded
When the page is loading, for a while the content gets shifted to up filling the logo position until the logo gets loaded. We need to fix this issue.

https://user-images.githubusercontent.com/78433013/148271672-a054373c-65c1-4f29-b571-93005c75d2e2.mp4

priyanshk20:
##### Closes #198 

This issue was occuring because, there was no height assigned to the image, I gave the image a height as well as wrapped it in a container.
- added `height` property to the image
- added `overflow-anchor: none;` and `z-index: 1;` to the image for further diminishing the issue.

### Is Development Tested? 
- [x] Yes
- [ ] No

Before: 

https://user-images.githubusercontent.com/52736997/148355615-1b71dc80-2b7c-44b1-9a68-2577af03541b.mp4

After:
bigScreen
smallScreen

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -108,7 +108,6 @@ body {
   height: 200px;
   display: block;
   margin: 20px auto;
-  overflow-anchor: none;
 }
 
 .social-media-links {
@@ -399,7 +398,6 @@ footer {
   .logo-img {
     width: 400px;
     height: 400px;
-    overflow-anchor: none;
   }
 
   .cta-button {
--- a/css/main.css
+++ b/css/main.css
@@ -103,11 +103,6 @@ body {
   font-weight: 700;
 }
 
-.img-container {
-  min-width: 200px;
-  min-height: 200px;
-}
-
 .logo-img {
   width: 200px;
   height: 200px;
--- a/css/main.css
+++ b/css/main.css
@@ -103,10 +103,18 @@ body {
   font-weight: 700;
 }
 
+.img-container {
+  min-width: 200px;
+  min-height: 200px;
+}
+
 .logo-img {
   width: 200px;
+  height: 200px;
   display: block;
   margin: 20px auto;
+  overflow-anchor: none;
+  z-index: 1;
 }
 
 .social-media-links {
@@ -396,6 +404,9 @@ footer {
 
   .logo-img {
     width: 400px;
+    height: 400px;
+    overflow-anchor: none;
+    z-index: 1;
   }
 
   .cta-button {
--- a/css/main.css
+++ b/css/main.css
@@ -109,7 +109,6 @@ body {
   display: block;
   margin: 20px auto;
   overflow-anchor: none;
-  z-index: 1;
 }
 
 .social-media-links {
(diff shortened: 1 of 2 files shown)


css/main.css
```diff
@@ -103,10 +103,18 @@ body {
   font-weight: 700;
 }
 
+.img-container {
+  min-width: 200px;
+  min-height: 200px;
+}
```
rohan09-raj, line 109: Can't it be done without this image container ??

rohan09-raj (commented):
Can you please test this by throttling the network ??
image
Refer to the Image and test this on `Fast 3G` and `Slow 3G`. Also, please do change it to `no throttling` when you are done with testing these otherwise you will feel how your internet got slow 😆 😉

css/main.css
priyanshk20: Let me test it quicky.

priyanshk20: Yes, the issue was solved without this image container, hence removed it.

css/main.css
```diff
@@ -105,8 +105,11 @@ body {
 
 .logo-img {
   width: 200px;
+  height: 200px;
   display: block;
   margin: 20px auto;
+  overflow-anchor: none;
+  z-index: 1;
```
rohan09-raj, line 112: Is z-index applying here ??
I think that it works only with `display: flex` and if we specify `position` attribute.

css/main.css
```diff
@@ -396,6 +399,9 @@ footer {
 
   .logo-img {
     width: 400px;
+    height: 400px;
+    overflow-anchor: none;
+    z-index: 1;
```
rohan09-raj, line 404: Here too refer to above comment

css/main.css
```diff
@@ -105,8 +105,11 @@ body {
 
 .logo-img {
   width: 200px;
+  height: 200px;
   display: block;
   margin: 20px auto;
+  overflow-anchor: none;
```
rohan09-raj, line 111: I don't think we need this, check what changes are there when this is not applied

css/main.css
```diff
@@ -396,6 +399,9 @@ footer {
 
   .logo-img {
     width: 400px;
+    height: 400px;
+    overflow-anchor: none;
```
rohan09-raj, line 403: Here too, same comment

priyanshk20: Yes, these lines are required, if deleted it will disturb the dimensions of the logo like this:
image

rohan09-raj: What if we only remove `z-index` ?? Will this be same in that case ??

priyanshk20: Removing `z-index` from the classes

css/main.css
priyanshk20: Yes, removed and checked

rohan09-raj (approved):
Approving the changes 🥳, great job @priyanshk20

Issue #219, opened 2022-01-17
Fix the styling of Github Sign in button 


gunjal-prajakt26:
![unknown](https://user-images.githubusercontent.com/82635500/149744336-2eb679a2-c4b3-4e04-adf5-c128d2c11830.png)
Need to make text and logo in a line .

gunjal-prajakt26:
assign @gunjal-prajakt26

PR #220, opened 2022-01-17
style: btn-login-text updated

gunjal-prajakt26:
### Issue
#219 issue is resolve

### What is the change?

-Done the styling of Github sign in button
-Make the text and logo in one line

### Is Development Tested?

-  Yes

Before :
unknown

After :
Untitled

mobile view :
Untitled1

### DIFF
--- a/css/navbar.css
+++ b/css/navbar.css
@@ -50,13 +50,18 @@ nav li a:hover {
   display: none;
 }
 
-.btn-login-text {
+.btn-login-text{
   border-radius: 6px;
   padding: 5px;
   background-color: #1d1283;
   color: white;
   cursor: pointer;
   border: 2px solid white;
+  display: flex;
+}
+
+.btn-login-text img{
+  padding-left: 0.3rem;
 }
 
 .user-greet {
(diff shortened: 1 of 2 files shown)


css/navbar.css
```diff
@@ -57,6 +57,11 @@ nav li a:hover {
   color: white;
   cursor: pointer;
   border: 2px solid white;
+  display: flex;
+}
+
+.btn-login-text img {
+  padding-left: 0.3rem;
```
rohan09-raj, line 64: Why added padding ?? Can you please send a screenshot if we don't add it ??

rohan09-raj: Thanks @shubham-y

rohan09-raj (approved):
Looks good to me 🎉 , approving the PR

shubham-y (commented):
There seems to be an issue with vertical alignment
image

We can  add padding at top
```css
.nav-login-li {
    padding-top: 2px;
}
```
Flex seems to be a better option, I guess, what say ??

shubham-y:
> Flex seems to be a better option, I guess, what say ??

@rohan09-raj I did not understand, where flex will be a better option

PR #242, opened 2022-02-23
fixing folder structure and  Refactor HTML code

aman7870:
Fixing folder structure


rohan09-raj:
The site is being migrated to EmberJS, therefore we don't need refactoring in the code.

PR #262, opened 2022-08-11
fixed user.js in case of no profile pic

### LINKED ISSUE #261: fetching an undefined resource
The console shows a 404 error for logged-in users without a profile picture.

RitikJaiswal75:
### Issue
If the user has no Profile picture the console throws a 404 error

### What is the change?

I changed the user.js to check if the profile image is present or not if yes render it else render the default profile logo

##### Explaination 
When I sign in using Github the profile pic is not provided so the navbar loads no image and searches for an undefined resource throwing an error in the console. 

- Done this: added 2 lines in user.js
- Made these changes: used if else case in lines 22 and 23

### Is Development Tested?

-  Yes

#### Before
image

#### After 
image


js/user.js
```diff
@@ -19,7 +19,8 @@ const setUserGreeting = (username, firstName, userProfilePicture) => {
     const userImgURL = userProfilePicture;
 
     userImgEl.forEach((element) => {
-      element.src = userImgURL;
+      if (!!userImgURL) element.src = userImgURL;
```
MehulKChaudhari, line 22: Why do `!!userImgURL`? simply `userImgURL` will work in the if-case, its readable and understandable

MehulKChaudhari: also, use proper `if(){} else{}` syntax its also easy to understand.

MehulKChaudhari (commented):
Add a few neat comments. Please do check

js/user.js
RitikJaiswal75: Ok will do that

RitikJaiswal75:
Made changes ap per the comments above

package.json
```diff
@@ -15,5 +15,8 @@
   },
   "pre-commit": [
     "check"
-  ]
+  ],
+  "dependencies": {
+    "local-ssl-proxy": "^1.3.0"
+  }
```
prakashchoudhary07, line 21: Should this be in dependencies?

prakashchoudhary07 (changes requested):

js/analytics.js
```diff
@@ -1 +1,43 @@
...
+      for (
+        void 0 !== a ? (u = e[a] = []) : (a = 'posthog'),
+          u.people = u.people || [],
+          u.toString = function (t) {
+            var e = 'posthog';
+            return 'posthog' !== a && (e += '.' + a), t || (e += ' (stub)'), e;
+          },
+          u.people.toString = function () {
+            return u.toString(1) + '.people (stub)';
+          },
```
prakashchoudhary07, line 28: Is this file changed? If not then please remove it.

prakashchoudhary07 (changes requested):

package.json
RitikJaiswal75: No that is just a local dependency just for testing

js/analytics.js
RitikJaiswal75: Ok

RitikJaiswal75: This is the same file but with improved formatting. It's automatically changed by `npm run check` without which I cant commit.

PR #263, opened 2022-08-11
Handling no profile pic for the user

RitikJaiswal75:
### Issue
- https://github.com/Real-Dev-Squad/website-www/issues/261

### What is the change?
- Added if condition in user.js for the profile picture src.

### Is Development Tested?

- [x] Yes
- [ ] No

### Before Change
image

### After Change
image

PR #265, opened 2022-08-13
Handling formatting for analytics.js

### LINKED ISSUE #264: Prevent library code from formatting
The library code "analytics.js" is formatted every time we run ```npm run fix``` which should not happen, if we do not run the command ```npm run fix``` the lint check fails and the workflow stops thus creating a staging/deployment problem.

RitikJaiswal75:
### Issue

The analytics.js gets formatted each time we run ```npm run fix``` which on skipping creates lint issues in the workflow 

### What is the change?

Added a prettier ignore file to skip check for analytics.js so as to prevent it from formatting

### Is Development Tested?

- [x] Yes
- [ ] No

### Before

image

### After

image

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+js/analytics.js


RitikJaiswal75:
Since #263 is already merged so this was closed. I will ensure to separate PR's for separate issues in future.

PR #284, opened 2022-09-12
adds map

### LINKED ISSUE #278: Add map to main site
Add map to main site which will show location of all the user of RDS using [Mapbox GL JS](https://docs.mapbox.com/mapbox-gl-js/guides/)

image

poonammohata:
### Issue
Closes #278 

### What is the change?
Added map, which shows RDS user's location, showing user image as a marker

### Is Development Tested?
- As of right now from the user's API we are not getting location details so for now, tested this with sample data, once the backend changes are complete, need to integrate and test with real data

### Short video of how this will look on the website
https://user-images.githubusercontent.com/54801934/189804189-ccc2f0fb-dce1-482d-8c6b-59b1e9c4a607.mov

### To do while deploying this code
- Need to add this in the build command
```cd js && echo -e "const MAPBOXGL_ACCESS_TOKEN = 'token';\n\nexport default MAPBOXGL_ACCESS_TOKEN;" > config.js```
- The above command will create a config.js file in the js folder, where the token will be defined and exported, which we have imported in the map.js file.
- In the above command in place of a token we need to add the token which we will get on Mapbox after creating a Mapbox account

**Note:**
- Right now this feature is flag based, need to pass `map=true` as query params in the URL, to show the map on the website.
- Markers on the map will be shown for only those users who have both location coordinates and image data.
- If in case no user has location coordinates and image data then- the map wouldn't be shown.

### Things need to be done, for this feature
- [ ] Once the backend changes are complete and we get location data, need to integrate with API and test it thoroughly
- [ ] Need to create an account on [Mapbox GL JS](https://docs.mapbox.com/mapbox-gl-js/guides/) for RDS
- [ ] Need to decide how much zoom will keep on the map, and what will be the center coordinates of the map, right now in the center coordinates I have given coordinates of India, and the zoom is 2.5

### Note to Reviewers: 
- the Percy generate snapshots test will fail as I have made UI changes. That cannot be fixed currently.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -695,16 +695,12 @@ footer {
   background-color: var(--color-vivid-pink);
 }
 
-#section-map {
-  display: none;
-}
-
 .map-title {
+  color: var(--color-text-secondary);
+  margin: 50px 20px 32px;
   text-align: center;
   font-weight: 700;
   font-size: 2rem;
-  color: var(--color-why-description-text);
-  margin: 50px 20px 32px;
   line-height: 150%;
 }
 
@@ -712,7 +708,7 @@ footer {
   height: 300px;
   overflow: hidden;
   border: 3px solid #e20665;
-  margin: 25px 10%;
+  margin: 2% 10%;
 }
 
 #map {
@@ -722,22 +718,23 @@ footer {
   width: 100%;
   height: 300px;
 }
-.mapboxgl-canvas-container {
-  width: 100% !important;
-}
+
 .mapboxgl-canvas {
   width: 100% !important;
 }
 
 .marker {
   display: block;
-  border: none;
   border-radius: 50%;
   cursor: pointer;
   padding: 0;
   width: 40px;
   height: 40px;
   background-size: 100%;
+}
+
+.element-display-remove {
+  display: none;
 }
 
 /* MEDIA QUERIES */
--- a/css/main.css
+++ b/css/main.css
@@ -707,7 +707,7 @@ footer {
 .map-box {
   height: 300px;
   overflow: hidden;
-  border: 3px solid #e20665;
+  border: 3px solid var(--color-events);
   margin: 2% 10%;
 }
 
--- a/css/main.css
+++ b/css/main.css
@@ -697,7 +697,7 @@ footer {
 
 .map-title {
   color: var(--color-text-secondary);
-  margin: 50px 20px 32px;
+  margin: 2em 1em 1em;
   text-align: center;
   font-weight: 700;
   font-size: 2rem;
@@ -716,11 +716,6 @@ footer {
   top: 0;
   bottom: 0;
   width: 100%;
-  height: 300px;
-}
-
-.mapboxgl-canvas {
-  width: 100% !important;
 }
 
 .marker {
--- a/css/main.css
+++ b/css/main.css
@@ -695,6 +695,51 @@ footer {
   background-color: var(--color-vivid-pink);
 }
 
+#section-map {
+  display: none;
+}
+
+.map-title {
+  text-align: center;
+  font-weight: 700;
+  font-size: 2rem;
+  color: var(--color-why-description-text);
+  margin: 50px 20px 32px;
+  line-height: 150%;
+}
+
+.map-box {
+  height: 300px;
+  overflow: hidden;
+  border: 3px solid #e20665;
+  margin: 25px 10%;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: css/main.css [hunk 1]
```diff
@@ -695,6 +695,48 @@ footer {
...
+  text-align: center;
+  font-weight: 700;
+  font-size: 2rem;
+  line-height: 150%;
+}
+
+.map-box {
+  height: 300px;
+  overflow: hidden;
+  border: 3px solid #e20665;
```
rohan09-raj, line 710: Can we please use CSS variables here?

css/main.css
```diff
@@ -695,6 +695,48 @@ footer {
...
+#map {
+  overflow: visible;
+  top: 0;
+  bottom: 0;
+  width: 100%;
+  height: 300px;
+}
+
+.mapboxgl-canvas {
+  width: 100% !important;
```
rohan09-raj, line 723: Is there any other to achieve the same without using `!important`?

js/map.js
```diff
@@ -0,0 +1,61 @@
...
+      );
+      addMarkersToMap(userHaveLocationAndImageDetails);
+    }
+  } catch (err) {
+    console.log('get users data error', err);
+  }
+};
+
+async function addMarkersToMap(userDetails) {
+  if (map._loaded && map._fullyLoaded && userDetails.length && showMap) {
```
rohan09-raj, line 43: Can we simplify these conditions by using reverse conditions means conditions with ! (not) operator?

js/map.js
```diff
@@ -0,0 +1,61 @@
...
+    userDetails.forEach(({ location_coordinates, picture }) => {
+      // Create a DOM element for each marker.
```
rohan09-raj, line 45: Line 46 is self explanatory, no need for this comment

js/map.js
```diff
@@ -0,0 +1,61 @@
...
+      const el = document.createElement('div');
```
rohan09-raj, line 46: Can we name it better please ?

js/map.js
```diff
@@ -0,0 +1,61 @@
...
+      el.className = 'marker';
+      el.style.backgroundImage = `url(${picture.url})`;
+      // Add markers to the map.
+      new mapboxgl.Marker(el).setLngLat(location_coordinates).addTo(map);
+      mapSection.classList.remove('element-display-remove');
+      map.resize();
+    });
+  } else {
+    return;
+  }
```
rohan09-raj, line 56: We can simplify this if else by using guard condition. Something like below
```
if (conditions) {
  return;
}

// rest of the code, here no need to put the if conditions
```

css/main.css
poonammohata: In the Mapbox library, in this canvas element, they have used inline CSS for width, so other than using !important, I didn't find anything to overwrite that style.

image

If you know, of any other way by which we can change the width, then please let me know.

css/main.css
```diff
@@ -695,6 +695,48 @@ footer {
...
+  border: 3px solid var(--color-events);
+  margin: 2% 10%;
+}
+
+#map {
+  overflow: visible;
+  top: 0;
+  bottom: 0;
+  width: 100%;
+  height: 300px;
```
rohan09-raj, line 719: Can we please make the units consistent here?

css/main.css
```diff
@@ -695,6 +695,48 @@ footer {
...
+  font-weight: 700;
+  font-size: 2rem;
+  line-height: 150%;
+}
+
+.map-box {
+  height: 300px;
+  overflow: hidden;
+  border: 3px solid var(--color-events);
+  margin: 2% 10%;
```
rohan09-raj, line 711: Can we please use pixels for margins and paddings?

css/main.css
rohan09-raj: Got it, thanks for explanation, resolving the conversation

rohan09-raj: Do Mapbox library not provide the CSS customization??
Also, why do you exactly want to change the width? What is the purpose here?

poonammohata: After updating css, not needed this anymore, removed it

css/main.css
rohan09-raj: Ahh ok, then % is fine I feel

rohan09-raj (approved):
LGTM 💯 Approving and merging the PR

Issue #301, opened 2022-10-31
Disable redirecting to `github.realdevsquad.com`  


prakashchoudhary07:
Disable redirecting to `github.realdevsquad.com` as `Deceptive site ahead` is being shown

This is being done here https://github.com/Real-Dev-Squad/website-www/blob/e73fbee0a16b059a5e8fd83a2851af6a930a94ec/js/index.js#L70-L74

prankurnov15:
I am assigning  this to @prankurnov15 (myself)

prankurnov15:
#308 is removed due to some technical issues on my machine raised the #309 as constant fix .

PR #308, opened 2022-11-03
Fixed : now can login with github directly

prankurnov15:
### Issue #301 

Changed the redirect URL.


prakashchoudhary07 (changes requested):
Since any packages are not being added, there is no point in updating package-lock.json
Please remove this.

prakashchoudhary07:
There is a button present for sign-in already which takes to signin button
https://github.com/Real-Dev-Squad/website-www/blob/ba3331d43a7da99495e64db9769142b66c397165/index.html#L92-L112

I think the code which updates sign-in button link to a fake GitHub page to disable

And that is happening here 
https://github.com/Real-Dev-Squad/website-www/blob/e73fbee0a16b059a5e8fd83a2851af6a930a94ec/js/index.js#L70-L74

What do you think?

prankurnov15:
I will remove the package.json

prankurnov15:
Yes you are right the button is responsible for redirecting the user to fake page , I have updated this to direct login with GitHub by updating the url inside the function.

prakashchoudhary07 (changes requested):
You have deleted the package-lock.json, It should have been reverted to the previous version before updating it in this PR

PR #318, opened 2022-11-23
Join Section | send selected data

RitikJaiswal75:
### Issue

to be merged only after: https://github.com/Real-Dev-Squad/website-backend/pull/835

The join section converts complete local storage to JSON and sends it to the backend via fetch call.
This can cause issues as some users might have some extra data in their local storage that might be sent along with the required data which is not accepted by the backend thus this PR serves the purpose to solve the issue

### DIFF
--- a/js/join.js
+++ b/js/join.js
@@ -97,7 +97,7 @@ const previewWhyRds = document.getElementById('previewWhyRds');
 const previewHeardAbout = document.getElementById('previewHeardAbout');
 const submit = document.getElementById('next4');
 
-//Variables for Completed page
+//Vatiables for Completed page
 const personalLink = document.getElementById('personalLink');
 const copyBtn = document.getElementById('copy');
 
--- a/js/join.js
+++ b/js/join.js
@@ -356,7 +356,6 @@ submit.addEventListener('click', async () => {
     })
     .catch((err) => {
       alert(`Error in saving user data ${err}`);
-      localStorage.setItem('flowState', 4);
     });
 });
 
--- a/js/join.js
+++ b/js/join.js
@@ -97,7 +97,7 @@ const previewWhyRds = document.getElementById('previewWhyRds');
 const previewHeardAbout = document.getElementById('previewHeardAbout');
 const submit = document.getElementById('next4');
 
-//Vatiables for Completed page
+//Variables for Completed page
 const personalLink = document.getElementById('personalLink');
 const copyBtn = document.getElementById('copy');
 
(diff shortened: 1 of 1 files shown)


js/join.js
```diff
@@ -314,11 +343,11 @@ submit.addEventListener('click', async () => {
     headers: {
       'Content-Type': 'application/json',
     },
-    body: data,
+    body: JSON.stringify(sendJoinData()),
   })
     .then((res) => res.json())
     .then((res) => {
-      if (res.statusCode !== 200) {
+      if (res.message != 'User Data Added Succesfully') {
```
RitikJaiswal75, line 350: statusCode is not being sent from backend

RitikJaiswal75: Will make changes in nest pr

js/join.js
```diff
@@ -243,6 +243,36 @@ function previewFiller() {
...
+    introduction,
+    skills,
+    college,
+    forFun,
+    whyRds,
+    foundFrom,
+    funFact,
+  };
+  return data;
+}
```
ankushdharkar, line 274: Could we generate it from a constants array please?
`['firstName', 'lastName', 'city', 'state'...]`

js/join.js
```diff
@@ -314,11 +343,11 @@ submit.addEventListener('click', async () => {
     headers: {
       'Content-Type': 'application/json',
     },
-    body: data,
+    body: JSON.stringify(sendJoinData()),
   })
     .then((res) => res.json())
     .then((res) => {
-      if (res.statusCode !== 200) {
```
ankushdharkar, line 321: Why are we getting rid of the status code here? This would name our error messaging inflexible and unmaintainable .

ankushdharkar (changes requested):

js/join.js
RitikJaiswal75: I used statusCode instead of `status` made the changes

js/join.js
```diff
@@ -106,14 +106,16 @@ function fetchSavedDetails() {
...
   })
-    .then((res) => res.json())
     .then((res) => {
-      window.localStorage.setItem('firstName', res.first_name);
-      window.localStorage.setItem('lastName', res.last_name);
       if (res.statusCode === 401) {
         alert('You are not logged in! Redirecting you to login.');
         location.href = GITHUB_OAUTH;
       }
+      res.json();
```
ankushdharkar, line 114: !

ankushdharkar (changes requested):

ankushdharkar (approved):

PR #329, opened 2022-11-30
[Ember-Migration] Created a header component

### LINKED ISSUE #320: [Ember-Migration] Create the header component
## Design Reference
image

rohan09-raj:
## Resolves Issue
Closes #320 

## What is the change?
- Created a responsive header 

## Is Development Tested?

- [x] Yes

### Web
image

### Tablet
image

### Mobile
image


heyrandhir (approved):
LGTM

app/components/header.hbs
```diff
@@ -0,0 +1,69 @@
...
+  <div class='header__container'>
+    <span class='container__text-social'>
+      <h2 class='text__subtitle'>
+        <span class='text__welcome'>Welcome</span>
+        <span class='text__to'>to the</span>
+      </h2>
+      <h1 class='text__title'>Real Dev Squad</h1>
+
+      <div class='social'>
+        <a href={{this.LINKEDIN_URL}} target='_blank' class='social__handles'>
```
harshith-venkatesh, line 17: Do you think we should be adding `noreferrer noopener` here https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

app/components/header.hbs
```diff
@@ -0,0 +1,69 @@
...
+        <span class='social__seperators-round'>&bull;</span>
+
+        <a href={{this.FACEBOOK_URL}} target='_blank' class='social__handles'>
+          Facebook
+          <img
+            class='social__icon'
+            src='assets/icons/facebook-logo.png'
+            alt='Facebook Icon'
+          />
+        </a>
```
harshith-venkatesh, line 48: Could we follow DRY(Don't Repeat Yourself) here?
Could we not look into having constant socialLinkProperties where we could have properties like href link, name of the link, image src and alt tag.
We could have one source of data there and could pass that in a function here

app/constants/urls.js
```diff
@@ -12,3 +12,10 @@ export const AUTH = {
   SIGN_IN:
     'https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97',
 };
+
+export const SOCIALS = {
+  LINKEDIN: 'https://www.linkedin.com/company/real-dev-squad',
+  TWITTER: 'https://twitter.com/realdevsquad?lang=en',
+  FACEBOOK: 'https://www.facebook.com/Real-Dev-Squad-108713777585062/',
+  INSTAGRAM: 'https://www.instagram.com/realdevsquad/?hl=en',
```
harshith-venkatesh, line 20: Something similar to this where you are also other other details in the object

harshith-venkatesh (changes requested):
Good PR @rohan09-raj , well descriptive and tracked PR.
I have requested for few changes, please look into it and in case of any doubts reach out

app/components/header.hbs
rohan09-raj: Yes, we should add that. Made the change

app/components/header.hbs
rohan09-raj: Yes, made the following change

harshith-venkatesh (approved):

Issue #281, opened 2022-09-06
Migration to Ember


rohan09-raj:
# What to do ?
- Create a separate branch named as `develop-ember`
- Delete all the code and start fresh with the ember code

# Progress

## PRs

PR #350, opened 2022-12-15
[Ember-Migration] Handled Sign In/ Sign Up Flow

rohan09-raj:
## Issue
#326 

## What is the change?
- Created a login service
- Added a loading behaviour for greetings
- Updated the navbar tests

## Is Development Tested?
- [x] Yes

## Note for the reviewer
- Tests for login service will be written in a separate PR tracked in #281


app/services/login.js
```diff
@@ -0,0 +1,31 @@
...
+      if (user) {
+        if (user.incompleteUserDetails) window.location.replace(AUTH.SIGN_UP);
+        this.isLoggedIn = true;
+        this.userData = user;
+      }
+    } catch (error) {
+      this.isLoggedIn = false;
+    } finally {
+      this.isLoading = false;
+    }
```
heyrandhir, line 29: Since the finally block will execute post try and catch block.

1. it will overwrite the isLoggedIn from true to false which we set in the try block. Is that the desired behavior?
2. In catch we can use console the error using `console.error`and we need not set `isLoggedIn` false as it will be set false in finally block.

rohan09-raj: `this.isLoggedIn` and `this.isLoading` are two different tracked properties.
- `this.isLoading` is set to false in `finally` block
- `this.isLoggedIn` is set to true in `try` and false in `catch` block

rohan09-raj: For the 2nd comment, I have added the `console.error`

heyrandhir: My Bad I confused `isLoading` and `isLoggedIn`. Right, in that case, it's fine what you have done.

heyrandhir (approved):
LGTM 🙂

package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
     "dotenv": "^16.0.2",
     "fastboot-app-server": "^3.3.2",
-    "node-fetch": "^3.2.10"
+    "node-fetch": "2"
```
harshith-venkatesh, line 28: Why is the version low here

harshith-venkatesh (approved):

package.json
rohan09-raj: We are using Fastboot and it requires CommonJS imports to be used therefore the stable and still running version of node-fetch with ESM as well as CommonJS support is version 2
https://github.com/node-fetch/node-fetch#loading-and-configuring-the-module
image

MehulKChaudhari (approved):

Issue #355, opened 2022-12-24
Missing footer in join page


ankushdharkar:
![RDS-Issue-missing-footer@2x](https://user-images.githubusercontent.com/1935403/209451247-d66c7499-655d-4701-b480-ada128a8ab4b.png)

sahsisunny:
bug has been fixed and also merged (PR #358 ) in develop branch

PR #358, opened 2022-12-27
footer position fixed

sahsisunny:
# Fix the footer position in join page

### Issue

#355

### What is the change?

Fix the footer position in join.css and add class name to main tag in join.html.

### Is Development Tested?

- [x] Yes
- [ ] No

Before :
image

After :
After

### DIFF
--- a/css/join.css
+++ b/css/join.css
@@ -1,5 +1,9 @@
 @import url('global-variables.css');
 
+.join-main {
+  min-height: 75vh;
+}
+
 .greeting {
   font: 700 24px 'Roboto', sans-serif;
   line-height: 28px;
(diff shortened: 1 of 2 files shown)


rohan09-raj (commented):
@sahsisunny 
In this following image, why the width is not covering the whole view port of the browser as can be seen in before screenshot.
Also, it seems that something is missing in the container of the `copy` button. Is that true?
image

sahsisunny:
> @sahsisunny In this following image, why the width is not covering the whole view port of the browser as can be seen in before screenshot. Also, it seems that something is missing in the container of the `copy` button. Is that true? image

Screenshot (25)

rohan09-raj (approved):

Issue #362, opened 2022-12-30
Dropdown click not working


satyam73:
I wanted to share one small issue in our main page(www-site) navigation bar. 
So in the navbar's dropdown my profile link click is only working when clicking on the text. In hovered area click is not working (image for reference).

![image](https://user-images.githubusercontent.com/85296770/210090004-d48578cf-33ce-4cc0-8b57-f6f43115a92d.png)

PS - For sign out it is working.

samyakshah3008:
I would love to work on this issue, can you please assign me?

Vaibhav2711:
Can you assign this issue to me

crater90:
can you assign this to me ?

prerana1821:
I would like to work on this issue.

satyam73:
> I would like to work on this issue.

Sorry @prerana1821 it's first come first serve basis so assigned to @crater90 .

PR #382, opened 2023-01-11
solved the dropdown issue

Vaibhav2711:
Issue
Dropdown click not working

Example : Closes #<issue_number>
Closes https://github.com/Real-Dev-Squad/website-www/issues/362

What is the change?
Wrapped the list item around anchor tag
added the missing id attribute to list tag in my profile
Provide a small description in points of what did you change

Example :
Done this
Made these changes
Is Development Tested?
[ ✅] Yes
 No
Before / After Change Screenshots

Example:
Before :
Image/Video

After :
Image/Video

Note :
What to delete and what not to

Keep the headings and answer the questions.
Delete the examples
Delete the description below headings What is the change? & Before/After Screenshots

### DIFF
--- a/index.html
+++ b/index.html
@@ -72,11 +72,9 @@
   <body>
     <div class="dropdown hide">
       <ul class="dropdown-list">
-        <li class="dropdown-item">
-          <a class="dropdown-link" href="https://my.realdevsquad.com/">
-            My Profile
-          </a>
-        </li>
+        <a class="dropdown-link" href="https://my.realdevsquad.com/">
+          <li class="dropdown-item" id="myprofile-option">My Profile</li>
+        </a>
         <hr class="line" />
         <li class="dropdown-item dropdown-link" id="signout-option">
           Sign out
--- a/index.html
+++ b/index.html
@@ -72,9 +72,11 @@
   <body>
     <div class="dropdown hide">
       <ul class="dropdown-list">
-        <a class="dropdown-link" href="https://my.realdevsquad.com/">
-          <li class="dropdown-item" id="myprofile-option">My Profile</li>
-        </a>
+        <li class="dropdown-item" id="myprofile-option">
+          <a class="dropdown-link" href="https://my.realdevsquad.com/">
+            My Profile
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -72,11 +72,9 @@
   <body>
     <div class="dropdown hide">
       <ul class="dropdown-list">
-        <li class="dropdown-item">
-          <a class="dropdown-link" href="https://my.realdevsquad.com/">
-            My Profile
-          </a>
-        </li>
+        <a class="dropdown-link" href="https://my.realdevsquad.com/">
```
satyam73, line 75: I think we can do this with the proper hierarchical way like ul > li > a , which also follows standards of HTML and will be good approach. Although we can do it like this but better approach will be good.

satyam73 (changes requested):

index.html
Vaibhav2711: Corrected as per requirements

sahsisunny: Satyam and Vaibhav can't we give it an approach like a `Sign Out` button?

satyam73: > 

Sunny your approach will not be good for accessibility I think because you are using li as anchor tag. So the screen readers and crawler will not be able to recognize it as a anchor rather it will be considered a li element.

sahsisunny: You are right, but the SIGNOUT button remains the same. So change the approach of the SIGNOUT button as well.

Vaibhav2711: Actually My Profile button can be changed to event handler like the SIGNOUT button. Also for SIGNOUT button we are making an api call first that's why anchor tag will have a problem, as it is not redirecting rather waiting for the api request to resolve and then reloading.

sahsisunny: It's not a `button` it is `li` right?

Vaibhav2711: Yes it's li elemenwnt but has an onclick event handler

Vaibhav2711: sort of working like a button

rohan09-raj:
The PR is stale and there is another PR which already resolved the linked issue #404

PR #392, opened 2023-01-17
Fix join section

sahsisunny:
### What is the change?

Following files are changed:

- `css/main.css` - Fix the description text and set text to center to anchor tag.
- `index.html` - Change button to anchor tag and add href(go to the info text section) and add id to the info text section.
- `js/index.js` - Remove unnecessary code of joinBtn button.

### Is Development Tested?

- [x] Yes
- [ ] No

After :

Fix Join Section

After Request Changes

Fix Line Height

### DIFF
--- a/index.html
+++ b/index.html
@@ -258,7 +258,9 @@
             put in any efforts for yourself. We want to value everyone’s time
             and efforts.
           </p>
-          <a class="btn-join" href="#info-note">Join the Squad</a>
+          <a class="btn-join" href="https://realdevsquad.com/join"
+            >Join the Squad</a
+          >
         </div>
       </section>
 
@@ -622,7 +624,7 @@
         <a href="./event-ssr-correctly-august-2020.html">SSR the right way</a>
       </div>
 
-      <h3 class="info-note" id="info-note">
+      <h3 class="info-note">
         To know how this page is developed,
         <br />
         join our Discord channel by contacting one of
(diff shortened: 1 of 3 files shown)


js/index.js
```diff
@@ -101,8 +101,3 @@ modalTriggers.forEach((trigger) => {
     });
   });
 });
-
-const joinBtn = document.getElementById('btn-join');
-joinBTN.addEventListener('click', () => {
-  window.location.href = 'https://realdevsquad.com/join';
-});
```
rohan09-raj, line 108: It's behind a feature flag, we can move ahead with the functionality of clicking the button
Check here :
https://realdevsquad.com/?join=true
https://realdevsquad.com

We only need to fix the functionality as mentioned in the issue, the click was not working

index.html
```diff
@@ -258,7 +258,7 @@ <h1 class="features-title">How can you join?</h1>
             put in any efforts for yourself. We want to value everyone’s time
             and efforts.
           </p>
-          <button class="btn-join" id="btn-join">Join the Squad</button>
+          <a class="btn-join" href="#info-note">Join the Squad</a>
```
rohan09-raj, line 261: This is not visible in the production until user puts the feature flag, so we can move ahead with the button

rohan09-raj (changes requested):
@sahsisunny Please look into UI, it still has too much whitespace between the elements and I think those two paragraphs can be combined to one as on desktop site, it looks odd to see this much line height.

js/index.js
sahsisunny: done

rohan09-raj (approved):

PR #408, opened 2023-02-24
add query param for join steps

### LINKED ISSUE #403: [Ember-Migration] Back button optimization with Query and Params
## Issue
- Currently, when we visit `/join` and move to other steps
- Then if we click browser `Back` button it goes to `/` route
- But the expectation here and a good UX is we should be directed to the previous page where user comes from.

## Suggested Solution
- We can use query params for e.g. -> `realdevsquad.com/join?step=1`
- Also, we can look for other solutions as well.

sahsisunny:
### What is the change?

1. `app/controllers/join.js` : This file contains the controller for the join route.
2. `app/components/stepper.js` : Add the query params to the stepper component.
3. `tests/unit/controllers/join-test.js` : Add the test for the join controller.
   - First we check `join` controller is exist or not.
   - Second we check `join` controller has expected the query params `step`.

### Is Development Tested?

- [ ] Yes
- [ ] No

### After Change Screenshots

https://user-images.githubusercontent.com/70854507/221082334-f5cf8095-f83a-42db-848a-fd36e37900a3.mp4

**Note:** We skip the test for the stepper component.

### DIFF
--- a/app/components/stepper.js
+++ b/app/components/stepper.js
@@ -11,7 +11,10 @@ export default class StepperComponent extends Component {
   @service toast;
   @tracked preValid = false;
   @tracked isValid = JSON.parse(localStorage.getItem('isValid')) ?? false;
-  @tracked currentStep = +localStorage.getItem('currentStep') ?? 0;
+  @tracked currentStep =
+    +new URLSearchParams(window.location.search).get('step') ??
+    +localStorage.getItem('currentStep') ??
+    0;
   TITLE_MESSAGES = TITLE_MESSAGES;
   @tracked stepOneData = JSON.parse(localStorage.getItem('stepOneData'));
   @tracked stepTwoData = JSON.parse(localStorage.getItem('stepTwoData'));
@@ -25,6 +28,7 @@ export default class StepperComponent extends Component {
     if (this.currentStep < 5) {
       this.currentStep += 1;
       localStorage.setItem('currentStep', this.currentStep);
+      window.history.pushState({}, '', `?step=${this.currentStep}`);
     }
   }
 
@@ -32,6 +36,7 @@ export default class StepperComponent extends Component {
     if (this.currentStep > 0) {
       this.currentStep -= 1;
       localStorage.setItem('currentStep', this.currentStep);
+      window.history.pushState({}, '', `?step=${this.currentStep}`);
     }
   }
 
--- a/app/components/stepper.js
+++ b/app/components/stepper.js
@@ -27,18 +27,18 @@ export default class StepperComponent extends Component {
 
   constructor() {
     super(...arguments);
-    this.router.transitionTo('join', {
-      queryParams: { step: localStorage.getItem('currentStep') },
-    });
+    window.onpopstate = () => {
+      this.currentStep = Number(
+        +new URLSearchParams(window.location.search).get('step')
+      );
+    };
   }
 
   @action incrementStep() {
     if (this.currentStep < 5) {
       this.currentStep += 1;
       localStorage.setItem('currentStep', this.currentStep);
-      this.router.transitionTo('join', {
-        queryParams: { step: this.currentStep },
-      });
+      this.router.transitionTo(`/join?step=${this.currentStep}`);
     }
   }
 
@@ -46,9 +46,7 @@ export default class StepperComponent extends Component {
     if (this.currentStep > 0) {
       this.currentStep -= 1;
       localStorage.setItem('currentStep', this.currentStep);
-      this.router.transitionTo('join', {
-        queryParams: { step: this.currentStep },
-      });
+      this.router.transitionTo(`/join?step=${this.currentStep}`);
     }
   }
 
--- a/app/components/stepper.js
+++ b/app/components/stepper.js
@@ -38,6 +38,7 @@ export default class StepperComponent extends Component {
         },
       });
     }
+    window.history.pushState({}, '', `?step=${this.currentStep}`);
   }
 
   @action incrementStep() {
--- a/app/components/stepper.js
+++ b/app/components/stepper.js
@@ -13,9 +13,7 @@ export default class StepperComponent extends Component {
   @tracked preValid = false;
   @tracked isValid = JSON.parse(localStorage.getItem('isValid')) ?? false;
   @tracked currentStep =
-    +localStorage.getItem('currentStep') ??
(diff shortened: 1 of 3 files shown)


tests/unit/controllers/join-test.js
```diff
@@ -0,0 +1,12 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'website-www/tests/helpers';
+
+module('Unit | Controller | join', function (hooks) {
+  setupTest(hooks);
+
+  // TODO: Replace this with your real tests.
```
shmbajaj: Resolve this TODO

shmbajaj: Also, add tests and please follow TDD>>

app/components/stepper.js
```diff
@@ -9,9 +9,13 @@ import { JOIN_URL } from '../constants/apis';
 export default class StepperComponent extends Component {
   @service login;
   @service toast;
+  @service router;
   @tracked preValid = false;
   @tracked isValid = JSON.parse(localStorage.getItem('isValid')) ?? false;
-  @tracked currentStep = +localStorage.getItem('currentStep') ?? 0;
+  @tracked currentStep =
+    +localStorage.getItem('currentStep') ??
+    +new URLSearchParams(window.location.search).get('step') ??
```
shmbajaj, line 17: Do we really need `window` or it can be done with `router-service`, please confirm.
Also check these docs RouterService method [currentRoute](https://api.emberjs.com/ember/4.10/classes/RouterService/properties/currentRoute?anchor=currentRoute&show=inherited%2Cprotected%2Cprivate) 

[RouteInfo](https://api.emberjs.com/ember/4.10/classes/RouteInfo)

app/components/stepper.js
```diff
@@ -21,17 +25,33 @@ export default class StepperComponent extends Component {
   setIsValid = (newVal) => (this.isValid = newVal);
   setIsPreValid = (newVal) => (this.preValid = newVal);
 
+  constructor() {
+    super(...arguments);
+    if (
+      localStorage.getItem('currentStep') <=
+      new URLSearchParams(window.location.search).get('step')
```
shmbajaj, line 32: Check above cooment.

app/components/stepper.js
```diff
@@ -21,17 +25,33 @@ export default class StepperComponent extends Component {
...
+        },
+      });
+    }
+  }
+
   @action incrementStep() {
     if (this.currentStep < 5) {
       this.currentStep += 1;
       localStorage.setItem('currentStep', this.currentStep);
+      window.history.pushState({}, '', `?step=${this.currentStep}`);
```
shmbajaj, line 47: Check the below snippet if it's good to go with as your using **Router** service
```jsx
    this.router.transitionTo({ queryParams: { paramName: 'paramValue' } });
 ```

File: app/components/stepper.js [hunk 5]
```diff
@@ -21,17 +25,33 @@ export default class StepperComponent extends Component {
...
     }
   }
 
   @action decrementStep() {
     if (this.currentStep > 0) {
       this.currentStep -= 1;
       localStorage.setItem('currentStep', this.currentStep);
+      window.history.pushState({}, '', `?step=${this.currentStep}`);
```
shmbajaj, line 55: Check previous comment.

shmbajaj (changes requested):
- Add Tests and Please follow TDD>>

tests/unit/controllers/join-test.js
sahsisunny: done

app/components/stepper.js
sahsisunny: done

app/components/stepper.js
sahsisunny: done

app/components/stepper.js
sahsisunny: done

tests/unit/controllers/join-test.js
```diff
@@ -0,0 +1,16 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'website-www/tests/helpers';
+
+module('Unit | Controller | join', function (hooks) {
```
shmbajaj, line 4: - Can you write a test for checking what is expected to render with step query param change?

- Also write a test for validating your logic for the rendering step, when using `localstorage` and other etc...

app/components/stepper.js
```diff
@@ -21,17 +25,28 @@ export default class StepperComponent extends Component {
   setIsValid = (newVal) => (this.isValid = newVal);
   setIsPreValid = (newVal) => (this.preValid = newVal);
 
+  constructor() {
+    super(...arguments);
+    window.onpopstate = () => {
+      this.currentStep = Number(
+        +new URLSearchParams(window.location.search).get('step')
```
shmbajaj, line 32: If it is possible to do the router service, to get a query param use that.
```js
    window.onpopstate = () => {
      const queryParams = this.router.currentRoute.queryParams;
      this.currentStep = Number(queryParams.step);
    };
```

app/components/stepper.js
shmbajaj, line 17: Use router service here to get query param.

shmbajaj (changes requested):
- Update Description for TDD

tests/unit/controllers/join-test.js
sahsisunny: We skip the test for the stepper component and for `join` controller see main comment(updated).

app/components/stepper.js
sahsisunny: not working, its always return 0

app/components/stepper.js
sahsisunny: now I am using `this.args.step` for taking the value from param.

sahsisunny: see [this](https://github.com/Real-Dev-Squad/website-www/pull/408#discussion_r1125305308)

tests/unit/controllers/join-test.js
shmbajaj: It's not for the stepper component, it's like what you expect when step query param changes.

app/components/stepper.js
shmbajaj: You can use `this.args.step` here instead of using `window`, correct if I am wrong.

shmbajaj (approved):
✅ LGTM.

rohan09-raj (approved):

PR #410, opened 2023-03-03
Feat wheel animations

prakashchoudhary07:
### What is the change?

Provide a small description in points of what did you change

##### Example :

- Done this
- Made these changes

### Is Development Tested?

- [ ] Yes
- [ ] No

##### Example:

Before :
Image/Video

After :
Image/Video

**Note :**
What to delete and what not to

- Keep the headings and answer the questions.
- Delete the examples
- Delete the description below headings `What is the change?` & `Before/After Screenshots`

### DIFF
--- a/app/components/members-data-container.js
+++ b/app/components/members-data-container.js
@@ -1,0 +1,20 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { service } from '@ember/service';
+
+export default class MembersDataContainerComponent extends Component {
+  @service store;
+
+  @tracked members = [];
+
+  @action async loadMembers() {
+    const data = await this.store
+      .query('user', {
+        size: 5,
+      })
+      .catch((err) => console.error(err));
+    console.log({ data });
+    this.members = data;
+  }
+}
--- a/app/components/wheel-animations.hbs
+++ b/app/components/wheel-animations.hbs
@@ -1,0 +1,3 @@
+    {{#each @users as |m|}}
+        d
+    {{/each}}
--- a/app/serializers/user.js
+++ b/app/serializers/user.js
@@ -1,0 +1,24 @@
+import ApplicationSerializer from './application';
+
+export default class UserSerializer extends ApplicationSerializer {
+  normalizeArrayResponse(store, primaryModelClass, payload) {
+    if ('error' in payload) {
+      const error = {
+        status: payload.statusCode,
+        title: payload.error,
+        details: payload.message,
+      };
+      return { error };
+    }
+    const data = payload.users.map((user) => {
+      const { id, ...other } = user;
+      return {
+        id,
+        type: primaryModelClass.modelName,
+        attributes: other,
+      };
+    });
+    const links = { ...payload.links, first: null, last: null };
+    return { data, links };
+  }
+}
--- a/tests/integration/components/members-data-container-test.js
+++ b/tests/integration/components/members-data-container-test.js
@@ -1,0 +1,26 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'website-www/tests/helpers';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | members-data-container', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    // Set any properties with this.set('myProperty', 'value');
+    // Handle any actions with this.set('myAction', function(val) { ... });
+
+    await render(hbs`<MembersDataContainer />`);
+
+    assert.dom(this.element).hasText('');
+
+    // Template block usage:
+    await render(hbs`
+      <MembersDataContainer>
+        template block text
+      </MembersDataContainer>
+    `);
+
+    assert.dom(this.element).hasText('template block text');
+  });
+});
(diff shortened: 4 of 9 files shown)


app/serializers/user.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+    const data = payload.users.map((user) => {
+      const { id, ...other } = user;
+      return {
+        id,
+        type: primaryModelClass.modelName,
+        attributes: other,
+      };
+    });
+    const links = { ...payload.links, first: null, last: null };
```
satyam73, line 21: what if `{ first: null,last: null }`

tests/integration/components/members-data-container-test.js
```diff
@@ -0,0 +1,80 @@
...
+module('Integration | Component | members-data-container', function (hooks) {
+  setupRenderingTest(hooks);
+
+  hooks.beforeEach(function () {
+    // Register the mock service
+    this.owner.register('service:store', MockService);
+  });
+
+  // TODO: Write this test after having some way to mock api calls
+  test('it renders', async function (assert) {
```
satyam73, line 34:
```suggestion
  test('MembersDataContainer renders', async function (assert) {
```

app/components/wheel-animations.hbs
```diff
@@ -0,0 +1,25 @@
+<section class='animation__container' data-test-wheel-animation-section>
+  {{#if @users}}
+    <div>
+      {{#each @users as |m index|}}
+        <a
+          href='https://members.realdevsquad.com/{{m.username}}'
+          target='_blank'
+          class='member__link'
+          rel='noopener noreferrer'
+          data-test-wheel-animation-member-link
```
satyam73, line 10: since you're mapping the @users you can use `data-test-wheel-animation-member-link={{somedynamicvalue}}`

app/components/wheel-animations.hbs
```diff
@@ -0,0 +1,25 @@
...
+        >
+          <img
+            style={{(this.memberStyle index)}}
+            class='circle member__img'
+            src='{{(this.resizeImageUrl m.picture.publicId 150 150)}}'
+            alt='{{m.first_name}} {{m.last_name}}'
+            data-test-wheel-animation-member-image
```
satyam73, line 17: `data-test-wheel-animation-member-image={{somedynamicvaluefrom<mobject>>}}`

app/styles/wheel-animations.css
```diff
@@ -0,0 +1,79 @@
...
+  background: #f6f7f8;
+  background-image: linear-gradient(
+    to right,
+    #f6f7f8 0%,
+    #edeef1 20%,
+    #f6f7f8 40%,
+    #f6f7f8 100%
+  );
+  background-repeat: no-repeat;
+  background-size: 150px 150px;
```
satyam73, line 11: can we use rem?

app/styles/wheel-animations.css
```diff
@@ -0,0 +1,79 @@
...
+  animation-timing-function: linear;
+}
+
+@keyframes placeholderShimmer {
+  0% {
+    background-position: -468px 0;
+  }
+
+  100% {
+    background-position: 468px 0;
```
satyam73, line 28: can we use rem here please?

app/components/wheel-animations.hbs
```diff
@@ -0,0 +1,25 @@
+<section class='animation__container' data-test-wheel-animation-section>
+  {{#if @users}}
+    <div>
+      {{#each @users as |m index|}}
+        <a
+          href='https://members.realdevsquad.com/{{m.username}}'
```
satyam73, line 6:
```suggestion
          href='https://members.realdevsquad.com/{{member.username}}'
```

satyam73: similarly please check other places

satyam73 (commented):
px can be rem I think

app/styles/wheel-animations.css
prakashchoudhary07: Should the size of the image change, when the default font size is changed

app/styles/wheel-animations.css
prakashchoudhary07: Should the size of the image change, when the default font size is changed

app/components/wheel-animations.hbs
prakashchoudhary07: I don't think there is any need for this, as when testing the content length and single elements links are checked to be enough, so there is no need to add a number, as there is no need to iterate over the same test

app/components/wheel-animations.hbs
prakashchoudhary07: I don't think there is any need for this, as when testing the content length and single elements links are checked to be enough, so there is no need to add a number, as there is no need to iterate over the same test

app/components/members-data-container.js
```diff
@@ -0,0 +1,72 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { service } from '@ember/service';
+
+const noOfMembersToShow = 5;
```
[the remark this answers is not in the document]
prakashchoudhary07: Didn't want to add this extra property to the class as it is a constant and used only in js file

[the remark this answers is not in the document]
satyam73: okay

app/serializers/user.js
satyam73: got it

tests/integration/components/members-data-container-test.js
satyam73: Yeah we can leave as it is

app/components/wheel-animations.hbs
satyam73: okay

satyam73 (approved):
LGTM

PR #440, opened 2023-03-18
Adding Number of hours input field in join section

### LINKED ISSUE #439: Updating the Join section.
### What is the issue?
Currently, join section has no field, where new member can input how many hours they will contribute in RDS.

### What is the solution?
Updating the join section to include - How many hours per week, a person will contribute.

### Changes in frontend:
-In this we need to add a new input field for hours, which will be number.(between 1-100),
-Add the user input number into local storage, so that the user can continue from where they left from.

### Changes in backend:
-Add a validator to validate the input number, make sure it is between 1-100.
-Make sure the preview button is only visible after all the validation checks have been successful.

FMK2312:
### What is the change?

Frontend:
-Added new input field with type number

Backend:
-Added the numberOfHours(new input) to localStorage
-Added the logic to show the preview of number of hours on preview page
-Added the numberOfHours to data object being sent to backend
-Added validation logic for numberOfHours

### Is Development Tested?

- [ ] Yes
- [ ] No

Why RDS Page changes:
joinsection_before
joinsection_after

Preview page changes:
joinsection_before2

joinsection_after2


FMK2312:
I had to add the validation logic for numberOfHours, because even after setting min=1 and max=100, a user can directly enter the desire value, which can be invalid. 
Also added the error message, so that the user can be notified why the form is not proceeding further.

js/join.js
```diff
@@ -173,21 +175,39 @@ function whyRdsPageValidator() {
...
     foundFrom.value != '' &&
-    whyRds.value.trim() != ''
+    whyRds.value.trim() != '' &&
+    numberOfHours.value != '' &&
+    parseInt(numberOfHours.value) > 1 &&
+    parseInt(numberOfHours.value) <= 100
   );
 }
 
 function dataValidator(element, size) {
```
TanishqSingla, line 185: Change the function name. 
E.g - `textFieldValidator`, `textValidation` etc

js/join.js
```diff
@@ -173,21 +175,39 @@ function whyRdsPageValidator() {
...
-  let counter = document.getElementById(element.id + 'Counter');
-  let words_left = size - element.value.trim().split(' ').length;
-  counter.innerText = `At least, ${words_left} more word(s) required`;
-  if (words_left <= 0) {
-    counter.innerText = '';
-  }
-  if (element.value.trim().split(' ').length >= size && element.value != '') {
-    element.classList.remove('incorrect-data');
+  if (element.id !== 'numberOfHours') {
```
TanishqSingla, line 186: Have a separate function for this input, Don't club it with the text input validators.

js/join.js
```diff
@@ -244,6 +264,7 @@ function previewFiller() {
   previewFunFact.innerHTML = window.localStorage.getItem('funFact');
   previewWhyRds.innerHTML = window.localStorage.getItem('whyRds');
   previewHeardAbout.innerHTML = window.localStorage.getItem('foundFrom');
+  previewNumberOfHours.innerHTML = window.localStorage.getItem('numberOfHours');
```
TanishqSingla, line 267: `innerHTML` would store the text with tags, E.g `<p><span>Hello</span></p>` here if we do `.innerHTML` on p tag it will return `<span>Hello</span>`. Are you sure we want this here, if you're only concerned with the text then consider using `textContent` or `innerText`

package-lock.json
```diff
@@ -1,222 +1,8 @@
 {
```
TanishqSingla, line 1: Why is this change here?

TanishqSingla (changes requested):

join.html
```diff
@@ -321,6 +321,25 @@ <h2 class="progress">Step 3 of 4: Why Real Dev Squad?</h2>
...
+          </label>
+          <br />
+          <input
+            type="number"
+            name="numberOfHours"
+            id="numberOfHours"
+            class="user-input input-regular"
+            min="1"
+            max="100"
+            step="1"
```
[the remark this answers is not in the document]
FMK2312: Ok. I will remove the step.

join.html
```diff
@@ -321,6 +321,25 @@ <h2 class="progress">Step 3 of 4: Why Real Dev Squad?</h2>
           ></textarea>
           <p id="whyRdsCounter" class="text-red counter"></p>
           <br />
+          <label class="input-label" for="numberOfHours"
+            >How many hours per week, are you willing to contribute?
+            <span class="text-red">*</span>
```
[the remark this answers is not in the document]
FMK2312: Actually, this is the way that was followed for every input field, so for the sake of consistency, i did the same.

js/join.js
FMK2312: ok, i will do that.

js/join.js
FMK2312: Ok.

package-lock.json
FMK2312: I dont remember altering this, let me look into it.

TanishqSingla: Can you please look into it

TanishqSingla (changes requested):

package-lock.json
FMK2312: Ok

FMK2312: i am having hard time identifying the root cause here, all i did was run the npm install, and it rewrites the package-lock.json, please tell if you have any idea as to why it is behaving like this @TanishqSingla

FMK2312: one more thing do we even need the packages object? 
It is mentioning the same things that we have inside the dependencies object, then why do we even need packages object?
@TanishqSingla

FMK2312: done @TanishqSingla

TanishqSingla (approved):

js/goto.js
```diff
@@ -17,7 +17,7 @@ function redirectionHandler(data) {
 }
 
 function showSignupFormIfIncomplete() {
-  fetch('https://api.realdevsquad.com/users/self', {
+  fetch('http://api.realdevsquad.com/users/self', {
```
sahsisunny, line 20: Why using `http` ?

FMK2312: Missed it. Very good catch 👁 @sahsisunny

RitikJaiswal75 (approved):
LGTM

TanishqSingla (approved):

PR #446, opened 2023-04-01
[Live-Site] Create the header component for live screen

### LINKED ISSUE #435: [Live-Site] Create the header component for live screen
image
> Component to design

Figma Link - https://www.figma.com/file/O8QLwPzTuH3RIWoYdxitln/Live-Page?node-id=0-1&t=Tm6ZyIzHAvbvrMGj-0

SanketDhabarde:
### What is the change?
Created the header for the live site 

### Is Development Tested?

- [x] Yes
- [ ] No

Before :
Screenshot (403)

After :

https://user-images.githubusercontent.com/40385118/232196176-a7b4d7b7-fed6-4429-8720-c6586f6a39be.mp4

Mobile View:

https://user-images.githubusercontent.com/40385118/233820802-d9700f64-86bf-4d89-b726-74dee676e9d7.mp4

### DIFF
--- a/app/components/live-header.hbs
+++ b/app/components/live-header.hbs
@@ -1,7 +1,7 @@
 <section class='live__header'>
   <Reusables::Button
     @variant='share'
-    @text='Screenshare'
+    @text={{if this.isSharing 'Stop sharing' 'Screenshare'}}
     @type='button'
     @onClick={{this.toggleShare}}
   />
--- a/app/components/live-header.hbs
+++ b/app/components/live-header.hbs
@@ -1,9 +1,5 @@
-<section data-test-live='header' class='live__header'>
-  <Reusables::Button
-    @variant='share'
-    @text={{if this.isSharing 'Stop sharing' 'Screenshare'}}
-    @type='button'
-    @test='share'
-    @onClick={{this.toggleShare}}
-  />
-</section>
+<Tabs class='live__header' {{on 'click' this.tabHandler}}>
+  {{#each this.TABS as |tab|}}
+    <Tab @label={{tab.label}} @variant={{if tab.active 'active' ''}} />
+  {{/each}}
+</Tabs>
--- a/app/components/live-header.hbs
+++ b/app/components/live-header.hbs
@@ -1,42 +1,8 @@
 <section class='live__header'>
-  <div class='live__header--left'>
-    <Reusables::Button
-      @variant='share'
-      @text='Screenshare'
-      @type='button'
-      @onClick={{this.clickHandler}}
-    />
-    <Reusables::Button
-      @variant='option'
-      @text='Survey'
-      @type='button'
-      @onClick={{this.clickHandler}}
-    />
-    <Reusables::Button
-      @variant='option'
-      @text='Questions'
-      @type='button'
-      @onClick={{this.clickHandler}}
-    />
-    <Reusables::Button
-      @variant='option'
-      @text='Previous events'
-      @type='button'
-      @onClick={{this.clickHandler}}
-    />
-    <Reusables::Button
-      @variant='option'
-      @text='Real Dev Squad'
-      @type='button'
-      @onClick={{this.clickHandler}}
-    />
-  </div>
-  <div class='live__header--right'>
-    <Reusables::Button
-      @variant='moderator'
-      @text='Moderator view'
-      @type='button'
-      @onClick={{this.clickHandler}}
-    />
-  </div>
+  <Reusables::Button
+    @variant='share'
+    @text='Screenshare'
+    @type='button'
+    @onClick={{this.toggleShare}}
+  />
 </section>
--- a/app/components/live-header.hbs
+++ b/app/components/live-header.hbs
@@ -1,13 +1,28 @@
-<Tabs
-  data-test-live='header'
-  class='live__header'
-  {{on 'click' this.tabHandler}}
->
-  {{#each this.TABS as |tab|}}
-    <Tab
-      @test={{tab.label}}
-      @label={{tab.label}}
-      @variant={{if tab.active 'active' ''}}
-    />
-  {{/each}}
+<Tabs data-test-tabs class='tabs'>
+  <section class='tabs_info'>
+    <button
+      data-test-toggle-button
+      type='button'
+      class='tabs_toggle'
+      {{on 'click' this.toggleTabs}}
+    >
+      <FaIcon
+        @icon={{if this.isTabOpen 'chevron-down' 'chevron-right'}}
+        @size='2x'
+      />
+    </button>
+    {{this.activeTab}}
+  </section>
+  <section
+    class='tabs_list {{if this.isTabOpen "active" ""}}'
+  >
+    {{#each this.TABS as |tab|}}
+      <Tab
+        @test={{tab.label}}
+        @label={{tab.label}}
+        @variant={{if tab.active 'active' ''}}
+        @onClick={{this.tabHandler}}
+      />
+    {{/each}}
+  </section>
 </Tabs>
--- a/app/components/live-header.hbs
+++ b/app/components/live-header.hbs
@@ -1,0 +1,24 @@
+<section class='live__header'>
+  <div class='live__header--left'>
+    <Reusables::Button @variant='share' @text='Screenshare' @type='button' />
+    <Reusables::Button @variant='option' @text='Survey' @type='button' />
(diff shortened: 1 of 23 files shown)


app/components/live-header.hbs
```diff
@@ -0,0 +1,8 @@
+<section class='live__header'>
+  <Reusables::Button
+    @variant='share'
```
satyam73, line 3: button variant can be only three according to reusable button
image

tests/acceptance/live-test.js
```diff
@@ -0,0 +1,14 @@
...
+import { setupApplicationTest } from 'website-www/tests/helpers';
+
+module('Acceptance | live', function (hooks) {
+  setupApplicationTest(hooks);
+
+  test('visiting /live', async function (assert) {
+    await visit('/live');
+
+    assert.strictEqual(currentURL(), '/live');
+    assert.dom('.btn-share').hasText('Screenshare');
```
satyam73, line 12: Ember provides us data-test selectors specifically for tests so I think we can use that here.

tests/integration/components/live-header-test.js
```diff
@@ -0,0 +1,26 @@
...
+import { render, click } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | live-header', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders the live header', async function (assert) {
+    await render(hbs`<LiveHeader />`);
+
+    assert.dom('.live__header').exists();
```
satyam73, line 12: Similarly we can also use here

tests/integration/components/live-header-test.js
```diff
@@ -0,0 +1,26 @@
...
+    assert.dom('.live__header .btn.btn-share').exists();
```
satyam73, line 13: similarly here and below.

satyam73 (changes requested):

app/components/live-header.hbs
SanketDhabarde: When we make components reusable, we don't restrict them. It will be like a component is reusable but you only need to use it in these given variants. These don't sound good to me. Please correct me if I am wrong.

tests/acceptance/live-test.js
SanketDhabarde: Okay, I can see we are using data-test everywhere in tests. I thought we use test id as the least priority selector in testing from [react-testing-library](https://testing-library.com/docs/queries/about/#priority). Anyways, thank you I will make the change.

prerana1821 (approved):
Great work! Looks good to me

app/styles/tab.module.css
```diff
@@ -0,0 +1,19 @@
+.tab {
+  border-radius: 9px;
+  width: 11rem;
+  height: 3rem;
+  color: var(--color-persian-indigo);
```
satyam73, line 5: I don't think we need extra color variables. 
The theme is already defined by the set of colors. 
So may be we can use that colors/

app/styles/tab.module.css
```diff
@@ -0,0 +1,19 @@
...
+  font-size: 1rem;
+  cursor: pointer;
+  margin: 10px 0px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.tab.active {
+  color: var(--color-crimson-red);
```
satyam73, line 16: This color is not matching with our theme. 
So maybe we should change this.

app/styles/variables.css
```diff
@@ -19,4 +19,8 @@
   --color-green: #49a82e;
 
   --color-light-navyblue: #1d128340;
+
+  --color-crimson-red: #9d0000;
```
satyam73, line 23: Similarly it will be removed from here.

app/components/live-header.hbs
```diff
@@ -0,0 +1,13 @@
+<Tabs
+  data-test-live='header'
+  class='live__header'
```
satyam73, line 3: Isn't is seems little confusive that we are adding class `live__header` to Tabs

app/components/live-header.hbs
```diff
@@ -0,0 +1,13 @@
+<Tabs
+  data-test-live='header'
```
satyam73, line 2: Similarly for data-test selector here

app/components/live-header.hbs
satyam73, line 1: Also please we can add change in text when moving to other tabs.
As right now only style is changing so please can we add conditional rendering of text(as of now, which will be later component that in place)

satyam73 (changes requested):

app/components/live-header.hbs
SanketDhabarde: We can keep `live__header-tab`. What do you think?

app/components/live-header.hbs
SanketDhabarde: I'm not sure what you mean here. There are a number of tabs and the selected one is the active tab. Also, we will have an entry which is the active tab.

SanketDhabarde:
Changed colors
image

app/components/live-header.hbs
satyam73: Like here in the basic tabs - https://mui.com/material-ui/react-tabs/ tabpanel is changing

app/components/live-header.hbs
satyam73: Let's keep this general so it can be reused in other things if needed in future.

.tab will be good.

SanketDhabarde: I was thinking generic in the sense that if someone wants to reuse tabs they can style accordingly to their need. But I got your point.

SanketDhabarde:
I have one suggestion on mobile view @satyam73. Instead of complementing style can we use the same styling of navbar on mobile view like this:

https://user-images.githubusercontent.com/40385118/233820802-d9700f64-86bf-4d89-b726-74dee676e9d7.mp4

satyam73:
> I have one suggestion on mobile view @satyam73. Instead of complementing style can we use the same styling of navbar on mobile view like this:
> 
>  mobile_tab.mp4

Yeah it looks good

app/components/live-header.hbs
SanketDhabarde: Added the Change in the tab panel. Based on the active tab the component will load.
https://user-images.githubusercontent.com/40385118/234921093-710b41a4-03c8-42be-af29-df13da24cbd1.mp4

app/styles/variables.css
```diff
@@ -19,4 +19,6 @@
   --color-green: #49a82e;
 
   --color-light-navyblue: #1d128340;
+
+  --color-cultured: #f6f6f6;
```
satyam73, line 23: I think we we can use variables which is prebuilt for instead of creating new one.

app/components/tab.hbs
satyam73, line 1: I think we can use Reusable button component which is already present.

app/styles/variables.css
satyam73, line 23: I think we can use already declared variable for this color

tests/integration/components/live-header-test.js
satyam73, line 1: I think one test case can also be written by passing all three props LiveHeader component.

satyam73 (changes requested):

app/styles/variables.css
SanketDhabarde: No colour matching the background with pink as active color. That's why used the color from UX given

app/components/tab.hbs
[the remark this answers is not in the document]
SanketDhabarde: The reusable button doesn't pass the param with onClick. I need to pass the selected tab.

SanketDhabarde: https://github.com/Real-Dev-Squad/website-www/pull/446#discussion_r1181050449

app/styles/variables.css
SanketDhabarde: https://github.com/Real-Dev-Squad/website-www/pull/446#discussion_r1181047022

tests/integration/components/live-header-test.js
SanketDhabarde: That won't be necessary. The other param is `activeTab`, which will load the component later. Currently, that is displaying the dummy text.

app/styles/variables.css
satyam73: Got it

rohan09-raj (approved):

PR #477, opened 2023-05-10
[Live-Site] create a modal for joining the live session

### LINKED ISSUE #463: [Live-Site] Create a modal for joining the live session
image
> Component to be created

Figma Link - https://www.figma.com/file/2VtPGJH61Qp0iarkKW6KTO/Live-Site-Detailed-Designs?node-id=0-1&t=mIdw0NLHXmNrU7Vz-0

SanketDhabarde:
### What is the change?

Added a component to join the Live. We will only collect the "Name", as the "Role" of the joined user will be determined by the link he/she joined with the queryParams.

### Is Development Tested?

- [x] Yes
- [ ] No

### Before / After Change

Before :
image

After :
https://github.com/Real-Dev-Squad/website-www/assets/40385118/6326f89d-73e9-436b-b75c-f76964b27bc8

Mobile view:

https://github.com/Real-Dev-Squad/website-www/assets/40385118/2efeba25-1a95-47f4-beec-7fd8dc028e32


app/controllers/live.js
```diff
@@ -19,6 +21,20 @@ export default class LiveController extends Controller {
...
+  @action inputHandler(e) {
+    this.name = e.target.value;
+  }
+
+  @action clickHandler(e) {
+    e.preventDefault();
+    //TODO: Add funtionality to join live session
+    if (this.name) {
+      this.isJoin = true;
+      console.log('Join as', this.name);
```
rohan09-raj, line 33: We can remove the console log here for now

app/components/live-join.hbs
```diff
@@ -0,0 +1,20 @@
+<section data-test-card class='card live__join'>
+  <h3 data-test-card-title class='card__title'>Join the Event</h3>
```
rohan09-raj, line 2: Can we remove the top bottom margin for this heading ?

app/components/live-join.hbs
```diff
@@ -0,0 +1,20 @@
...
+  <form onsubmit={{@clickHandler}}>
+    <Reusables::InputBox
+      @field='Name'
+      @name='name'
+      @type='text'
+      @required={{true}}
+      @value={{@name}}
+      @onInput={{@inputHandler}}
+    />
```
rohan09-raj, line 11: Can we please give a placeholder here something like `Enter your name`

app/controllers/live.js
```diff
@@ -11,6 +11,8 @@ export default class LiveController extends Controller {
   ];
   @tracked activeTab = 'Screenshare';
   @tracked isLoading = true;
+  @tracked name = '';
+  @tracked isJoin = false;
```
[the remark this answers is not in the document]
SanketDhabarde: Yes, that seems better. I will update it.

app/components/live-join.hbs
SanketDhabarde: Yes sure. I was using an existing class so didn't mess with it.

SanketDhabarde:
New Card
image

rohan09-raj (approved):

satyam73:
Hey @SanketDhabarde 

Wouldn't the name live-join seems confusing?
How about renaming this to JoinLiveModal which is inherited by the generic modal?

Also the modal input box is taking too much width which is not looking good and the label in centered is not suiting also.

Can we place label to start of the input box?
Also try looking into the width of the input box.

SanketDhabarde:
> Hey @SanketDhabarde
> 
> Wouldn't the name live-join seems confusing? How about renaming this to JoinLiveModal which is inherited by the generic modal?
> 
> Also the modal input box is taking too much width which is not looking good and the label in centered is not suiting also.
> 
> Can we place label to start of the input box? Also try looking into the width of the input box.

Hi @satyam73. There is no generic modal. It's actually a generic card component classes. The reason for giving it as live-join is, first there is already a join section so I don't want it to be confused with it. And second to keep consistency between our components. Like live-header, live-join, etc. 

As it was a generic card the styling was like that to keep the consistency with other parts of forms in the application. If you think it's not good we need a small input box and the label at the start I can do that no worries but we have to compromise on the consistency. Let me know what you think about it.

satyam73:
> > Hey @SanketDhabarde
> > Wouldn't the name live-join seems confusing? How about renaming this to JoinLiveModal which is inherited by the generic modal?
> > Also the modal input box is taking too much width which is not looking good and the label in centered is not suiting also.
> > Can we place label to start of the input box? Also try looking into the width of the input box.
> 
> Hi @satyam73. There is no generic modal. It's actually a generic card component classes. The reason for giving it as live-join is, first there is already a join section so I don't want it to be confused with it. And second to keep consistency between our components. Like live-header, live-join, etc.
> 
> As it was a generic card the styling was like that to keep the consistency with other parts of forms in the application. If you think it's not good we need a small input box and the label at the start I can do that no worries but we have to compromise on the consistency. Let me know what you think about it.

Let's create a new generic modal and use that in our live-join modal also live-join-modal will be good because overall it's a modal. And if there comes a need of modal in future that can also be reused.

I would say take the card properties you want by adding that class(`card`)  and use another class to make it like we want. 
Since this is not a card so I don't think there will be any consistency issue.

Also try putting the label to the start. Something like in the image below:
image

SanketDhabarde:
> > > Hey @SanketDhabarde
> > > Wouldn't the name live-join seems confusing? How about renaming this to JoinLiveModal which is inherited by the generic modal?
> > > Also the modal input box is taking too much width which is not looking good and the label in centered is not suiting also.
> > > Can we place label to start of the input box? Also try looking into the width of the input box.
> > 
> > 
> > Hi @satyam73. There is no generic modal. It's actually a generic card component classes. The reason for giving it as live-join is, first there is already a join section so I don't want it to be confused with it. And second to keep consistency between our components. Like live-header, live-join, etc.
> > As it was a generic card the styling was like that to keep the consistency with other parts of forms in the application. If you think it's not good we need a small input box and the label at the start I can do that no worries but we have to compromise on the consistency. Let me know what you think about it.
> 
> Let's create a new generic modal and use that in our live-join modal also live-join-modal will be good because overall it's a modal. And if there comes a need of modal in future that can also be reused.
> 
> I would say take the card properties you want by adding that class(`card`) and use another class to make it like we want. Since this is not a card so I don't think there will be any consistency issue.
> 
> Also try putting the label to the start. Something like in the image below: image

With the label at the start, it will look good if we have other inputs as well. With only one input, It's not looking good. here's how it will look:
**Ex with the label at the start:**
name_start
**Ex with the small input:**
image

And I don't think it's modal. Modal is something subordinate to the main screen, not the first screen visible. And this join section will be visible first when you are on the live page. [Wiki](https://en.wikipedia.org/wiki/Modal_window).

Let me know what you think.

satyam73 (approved):
LGTM👍

PR #484, opened 2023-05-19
fix intro page 

### LINKED ISSUE #483: Fix Intro page
## Issue

- When user try to access  https://realdevsquad.com/intro?id=ZDlUE3meJV3XzVyc1yI7 before login or Sign In With GitHub then the page looks like below image
image

- As a presenter, Our intro page seems to be broken.
- As a user, This makes user confused

## Solution
This is 401 error So we have to show 401 error on page and also alert user you are not log-in and redirect to login page.

shubhamsigdar1:
### What is the change?

- In intro.js file, added 401 image into page and alert message when user is not login.

### Is Development Tested?

- [x] Yes
- [ ] No

### Before Image

image

### After change video

https://github.com/Real-Dev-Squad/website-www/assets/107163260/2201405d-89e3-4ab3-8e05-02b1387903e7

### DIFF
--- a/js/intro.js
+++ b/js/intro.js
@@ -295,9 +295,11 @@ async function showSavedDetails() {
     }
   } catch (err) {
     console.error(err);
-    window.addEventListener('click', showToast('something went wrong', 6000));
+    loading.classList.add('hidden');
+    generatenotAuthorizedPage();
     setTimeout(function () {
+      alert('You are not logged in! Redirecting you to login.');
       window.location.href = HOME_URL;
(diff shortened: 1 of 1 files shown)


RitikJaiswal75 (approved):
Tested it locally, it is working, Great and fast work @shubhamsigdar1 
Thankyou

harshith-venkatesh (approved):
Approving it as it is prod bug, I believe this is thoroughly tested in local and staging

PR #494, opened 2023-06-06
Bypassing word limit

vaishalimaheshwari2000:
The text input fields in https://realdevsquad.com/join require users to fill in a few details.
Amongst them, there are fields that require users to fill in a minimum of 100 words.
But, the users can easily bypass that check by using multiple spaces and then putting a character
or by using multiple special characters separated by a space.

https://github.com/Real-Dev-Squad/website-www/assets/80218238/fc117e98-cef3-493e-b0c3-c25f09718cc3

### DIFF
--- a/js/join.js
+++ b/js/join.js
@@ -185,50 +185,85 @@ function arePersonalDetailsValid() {
 console.log(introduction.value);
 function introPageValidator() {
   let countIntroduction = 0;
-  let textAreaIntroduction = Array.from(
-    document.querySelector('#introduction').value,
-  );
-  for (let i = 0; i < textAreaIntroduction.length; i++) {
-    let char = textAreaIntroduction[i];
-    if (/S+/.test(char)) {
+  let textAreaIntroduction = document.querySelector('#introduction').value;
+
+  let m;
+  const regex = /\w+/g;
+
+  while ((m = regex.exec(textAreaIntroduction)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
+    if (m.index === regex.lastIndex) {
+      regex.lastIndex++;
+    }
+
+    // The result can be accessed through the `m`-variable.
+    m.forEach((match, groupIndex) => {
       countIntroduction++;
-    }
+    });
   }
 
-  let textAreaSkills = Array.from(document.querySelector('#skills').value);
+  let textAreaSkills = document.querySelector('#skills').value;
   let countSkills = 0;
-  for (let i = 0; i < textAreaSkills.length; i++) {
-    let char = textAreaSkills[i];
-    if (/S+/.test(char)) {
+  let m2;
+
+  while ((m2 = regex.exec(textAreaSkills)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
+    if (m2.index === regex.lastIndex) {
+      regex.lastIndex++;
+    }
+
+    // The result can be accessed through the `m`-variable.
+    m2.forEach((match, groupIndex) => {
       countSkills++;
-    }
+    });
   }
 
-  let textAreaCollege = Array.from(document.querySelector('#college').value);
+  let textAreaCollege = document.querySelector('#college').value;
   let countCollege = 0;
-  for (let i = 0; i < textAreaCollege.length; i++) {
-    let char = textAreaCollege[i];
-    if (/S+/.test(char)) {
+  let m3;
+  while ((m3 = regex.exec(textAreaCollege)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
+    if (m3.index === regex.lastIndex) {
+      regex.lastIndex++;
+    }
+
+    // The result can be accessed through the `m`-variable.
+    m3.forEach((match, groupIndex) => {
       countCollege++;
-    }
+    });
   }
 
-  let textAreaForFun = Array.from(document.querySelector('#forFun').value);
+  let textAreaForFun = document.querySelector('#forFun').value;
   let countForFun = 0;
-  for (let i = 0; i < textAreaForFun.length; i++) {
-    let char = textAreaForFun[i];
-    if (/S+/.test(char)) {
-      countForFun++;
+  let m4;
+
+  while ((m4 = regex.exec(textAreaForFun)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
+    if (m4.index === regex.lastIndex) {
+      regex.lastIndex++;
     }
+
+    // The result can be accessed through the `m`-variable.
+    m4.forEach((match, groupIndex) => {
+      countForFun++;
+    });
   }
 
-  let textAreafunFact = Array.from(document.querySelector('#funFact').value);
+  let textAreafunFact = document.querySelector('#funFact').value;
   let countfunFact = 0;
-  for (let i = 0; i < textAreafunFact.length; i++) {
-    let char = textAreafunFact[i];
-    if (/S+/.test(char)) {
-      countfunFact++;
+
+  let mFun;
+
+  while ((mFun = regex.exec(textAreafunFact)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
+    if (mFun.index === regex.lastIndex) {
+      regex.lastIndex++;
     }
+
+    // The result can be accessed through the `m`-variable.
+    mFun.forEach((match, groupIndex) => {
+      countfunFact++;
+    });
   }
 
   return (
--- a/js/join.js
+++ b/js/join.js
@@ -182,7 +182,6 @@ function arePersonalDetailsValid() {
     country.value.trim() != ''
   );
 }
-console.log(introduction.value);
 function introPageValidator() {
   let countIntroduction = 0;
   let textAreaIntroduction = document.querySelector('#introduction').value;
@@ -473,9 +472,7 @@ copyBtn.addEventListener('click', () => {
   navigator.clipboard.writeText(`Real Dev Squad Verification Link: ${url}`);
 });
 
-document
-  .querySelector('#introduction')
-  .addEventListener('mouseleave', (event) => {
+introduction.addEventListener('mouseleave', (event) => {
     let count = 0;
     let textArea = document.querySelector('#introduction').value;
     let m;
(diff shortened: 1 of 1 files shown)


js/join.js
```diff
@@ -225,16 +225,19 @@ function numberDataValidator(element, min, max) {
...
+  // console.log(element.length);
+  if (size != 0) counter.innerText = `At least, ${size} more word(s) required`;
+  // if (size <= 0 && element.value!='') {
+  //   counter.innerText = ' ';
+  // }
+  // if (size && element.value != '') {
+  // element.classList.remove('incorrect-data');
+  // } else {
+  if (size != 0) element.classList.add('incorrect-data');
+  // }
```
RitikJaiswal75, line 240: please remove the unused commented code

js/join.js
```diff
@@ -410,3 +413,95 @@ submit.addEventListener('click', async () => {
 copyBtn.addEventListener('click', () => {
   navigator.clipboard.writeText(`Real Dev Squad Verification Link: ${url}`);
 });
+
+document.querySelector('#introduction').addEventListener('mouse', (event) => {
+  let count = 0;
+  let textArea = Array.from(document.querySelector('#introduction').value);
+  // console.log(textArea.length);
```
RitikJaiswal75, line 420: please remove this line

RitikJaiswal75 (changes requested):
Please remove all the commented code and Provide a proper description for the pr and link the issue ticket to the pr.

js/join.js
```diff
@@ -410,3 +413,95 @@ submit.addEventListener('click', async () => {
 copyBtn.addEventListener('click', () => {
   navigator.clipboard.writeText(`Real Dev Squad Verification Link: ${url}`);
 });
+
+document.querySelector('#introduction').addEventListener('mouse', (event) => {
```
shubhamsigdar1, line 417: why are you using mouse event??

shubhamsigdar1 (commented):
Hello @vaishalimaheshwari2000 Can you please add one video in description where you are checking all the test cases

js/join.js
vaishalimaheshwari2000: Okay, I'll do

js/join.js
vaishalimaheshwari2000: Okay, I'll do

js/join.js
vaishalimaheshwari2000: Done, video is attached in the description. Could you please look into it. Thank you

js/join.js
```diff
@@ -182,18 +182,65 @@ function arePersonalDetailsValid() {
     country.value.trim() != ''
   );
 }
-
+console.log(introduction.value);
```
sahsisunny, line 185:
```suggestion
```

js/join.js
```diff
@@ -410,3 +449,133 @@ submit.addEventListener('click', async () => {
...
+    // The result can be accessed through the `m`-variable.
+    m.forEach((match, groupIndex) => {
+      count++;
+    });
+  }
+
+  if (count < 5) {
+    dataValidator(document.querySelector('#skills'), 5 - count);
+  } else if (count >= 5) {
+    document.querySelector('#skillsCounter').innerText = '';
```
sahsisunny, line 502: For this type of statement please create variables the use it

js/join.js
```diff
@@ -410,3 +449,133 @@ submit.addEventListener('click', async () => {
...
+  let m;
+  const regex = /\w+/g;
+
+  while ((m = regex.exec(textArea)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
+    if (m.index === regex.lastIndex) {
+      regex.lastIndex++;
+    }
+
+    // The result can be accessed through the `m`-variable.
```
sahsisunny, line 543: If the comments are not so important can you please remove it ?

js/join.js
vaishalimaheshwari2000: Done

js/join.js
vaishalimaheshwari2000: Done

js/join.js
```diff
@@ -410,3 +449,133 @@ submit.addEventListener('click', async () => {
...
+  }
+});
+document.querySelector('#college').addEventListener('mouseleave', (event) => {
+  let textArea = document.querySelector('#college').value;
+  let count = 0;
+  let m;
+  const regex = /\w+/g;
+
+  while ((m = regex.exec(textArea)) !== null) {
+    // This is necessary to avoid infinite loops with zero-width matches
```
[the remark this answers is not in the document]
vaishalimaheshwari2000: Yes, because it finds a match , if it finds. It returns an array. Otherwise returns null

js/join.js
vaishalimaheshwari2000: Done

js/join.js
```diff
@@ -182,18 +182,65 @@ function arePersonalDetailsValid() {
...
+
+  let textAreafunFact = Array.from(document.querySelector('#funFact').value);
+  let countfunFact = 0;
+  for (let i = 0; i < textAreafunFact.length; i++) {
+    let char = textAreafunFact[i];
+    if (/S+/.test(char)) {
+      countfunFact++;
+    }
+  }
+
```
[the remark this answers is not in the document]
vaishalimaheshwari2000: This can't be refactor according to my perspective. Because when user filled correctly data, then the nextbutton will enable, before that this method checks the data validation

js/join.js
```diff
@@ -410,3 +471,119 @@ submit.addEventListener('click', async () => {
...
+  while ((m = regex.exec(textArea)) !== null) {
+    if (m.index === regex.lastIndex) {
+      regex.lastIndex++;
+    }
+
+    m.forEach((match, groupIndex) => {
+      count++;
+    });
+  }
```
khushi818, line 535: please refactor the code which are repetitive. Make a function for it.

khushi818 (changes requested):

js/join.js
```diff
@@ -410,3 +424,81 @@ submit.addEventListener('click', async () => {
...
+  return count;
+}
+
+introduction.addEventListener('mouseleave', (event) => {
```
shubhamsigdar1, line 444: Nit: why you used mouse leave event listener??

js/join.js
```diff
@@ -410,3 +424,81 @@ submit.addEventListener('click', async () => {
 copyBtn.addEventListener('click', () => {
   navigator.clipboard.writeText(`Real Dev Squad Verification Link: ${url}`);
 });
+
+function countWords(textArea) {
+  let m;
+  const regex = /\w+/g;
+  let count = 0;
+  while ((m = regex.exec(textArea)) !== null) {
```
shubhamsigdar1, line 432: what is the meaning of this line??
what is the meaning of m??

js/join.js
```diff
@@ -410,3 +424,81 @@ submit.addEventListener('click', async () => {
...
+  let textArea = funFact.value;
+  let count = 0;
+  count = countWords(textArea);
+  if (count < 100) {
+    dataValidator(funFact, 100 - count);
+  } else if (count >= 100) {
+    document.querySelector('#funFactCounter').innerText = '';
+    funFact.classList.remove('incorrect-data');
+  }
+});
```
shubhamsigdar1, line 504: you are writing the same code steps 5times...just changing the textArea name...can we optimize 5steps into 1steps

js/join.js
```diff
@@ -182,18 +182,33 @@ function arePersonalDetailsValid() {
...
+  let textAreaIntroduction = introduction.value;
+  let textAreaSkills = skills.value;
+  let countSkills = 0;
+  let textAreaCollege = college.value;
+  let countCollege = 0;
+  let textAreaForFun = forFun.value;
+  let countForFun = 0;
+  let textAreafunFact = funFact.value;
+  let countfunFact = 0;
+  countIntroduction = countWords(textAreaIntroduction);
```
shubhamsigdar1, line 196: want to know you are using countWords function here(i.e line- 196) and on line 446 also?
what is the difference between both??

js/join.js
```diff
@@ -182,18 +182,33 @@ function arePersonalDetailsValid() {
     country.value.trim() != ''
   );
 }
-
 function introPageValidator() {
+  let countIntroduction = 0;
```
shubhamsigdar1, line 186: You are correcting textarea of the intro page only not for other page

shubhamsigdar1 (commented):
Great work @vaishalimaheshwari2000 please check some comment

js/join.js
vaishalimaheshwari2000: In this both function has same working, here, in 196 I'm using because when user fill correctly data .Then, next button enable .
And in 446, it check the data validation. If  the data is incorrect then red box which appear , and apart from that if data is correct then. Red-box shouldn't be appear on the textarea field.

js/join.js
vaishalimaheshwari2000: In which ` while ((m = regex.exec(textArea)) !== null)` m is declared as a variable.
regex is defined as a regular expression pattern /\w+/g. This pattern matches one or more word characters (letters, digits, or underscores) in a global search.
m = regex.exec(textArea): This line executes the regular expression regex on the textArea and assigns the result to the variable m. The exec() method is used to search for matches of the regular expression pattern in the textArea.
(m = regex.exec(textArea)) !== null: The condition (m = regex.exec(textArea)) !== null checks if the result of the exec() method is not null. If null, it means that there are no more matches found in the textArea, and the loop will stop executing.

js/join.js
vaishalimaheshwari2000: I think. It's not possible because ecah textArea has their own scope. They're exist in different different function

js/join.js
vaishalimaheshwari2000: The mouseleave event is triggered when the mouse pointer leaves that element and also user can also get idea from that . How much words are left. They don't need to be count manually .

js/join.js
vaishalimaheshwari2000: Because this function returns a boolean value and like on line `263` it used in a if condition, where it decides that whether in the button , the button-disabled class  should be remove  and button-enabled class should be add . It depends on the introPageValidator function. And this fiunction only works on introPageValidator . Because it only tells about the next Button condition

PR #498, opened 2023-06-18
[Live Site] Add backend API calls

### LINKED ISSUE #491: [Live-Site] Add the Join functionality in live share
### AC
- Guest and Host  users will be able to join the live share just by putting their name
- The role of the user will be determined from the query param (role)
- Integrate 100ms SDK to join the dummy room which will be replaced with backed API in future

### UX
https://www.figma.com/file/2VtPGJH61Qp0iarkKW6KTO/Live-Site-Detailed-Designs?type=design&node-id=0-1&t=aWR2yCHfQdp1BaF4-0

SanketDhabarde:
### What is the change?
Previously we were joining the live screen share with the 100ms API. Now we are using our own Backend APIs to create and enter the room

### Is Development Tested?

- [x] Yes
- [ ] No

### Video

https://github.com/Real-Dev-Squad/website-www/assets/40385118/9114a622-f2d2-4b14-8004-9a1c3772cb78

### DIFF
--- a/app/constants/live.js
+++ b/app/constants/live.js
@@ -1,7 +1,6 @@
 export const ROLES = {
   guest: 'guest',
   host: 'host',
-  presenter: 'presenter',
 };
 
 export const API_METHOD = { POST: 'POST', GET: 'GET', PATCH: 'PATCH' };
--- a/app/constants/live.js
+++ b/app/constants/live.js
@@ -1,3 +1,7 @@
-export const ROLES = { guest: 'guest', host: 'host', presenter: 'presenter' };
+export const ROLES = {
+  guest: 'guest',
+  host: 'host',
+  presenter: 'presenter',
+};
 
-export const API_METHOD = { POST: 'POST', GET: 'GET' };
+export const API_METHOD = { POST: 'POST', GET: 'GET', PATCH: 'PATCH' };
--- a/app/controllers/live.js
+++ b/app/controllers/live.js
@@ -28,7 +28,7 @@ export default class LiveController extends Controller {
     super(...arguments);
     setTimeout(() => {
       this.isLoading = false;
-    }, 5000);
+    }, 4000);
   }
 
   @action inputHandler(e) {
--- a/app/controllers/live.js
+++ b/app/controllers/live.js
@@ -17,7 +17,6 @@ export default class LiveController extends Controller {
   @tracked activeTab = 'Screenshare';
   @tracked isLoading = true;
   @tracked name = '';
-  @tracked isJoined = false;
   @tracked role = null;
   @globalRef('videoEl') videoEl;
   get liveService() {
@@ -28,7 +27,7 @@ export default class LiveController extends Controller {
     super(...arguments);
     setTimeout(() => {
       this.isLoading = false;
-    }, 4000);
+    }, 5000);
   }
 
   @action inputHandler(e) {
@@ -41,7 +40,6 @@ export default class LiveController extends Controller {
     const isHost = this.role === ROLES.host;
     if (this.name && (isGuest || isHost)) {
       this.liveService.joinSession(this.name, this.role);
-      this.isJoined = true;
       this.name = '';
     }
   }
@@ -55,7 +53,6 @@ export default class LiveController extends Controller {
 
   @action leaveSession() {
     this.liveService.leaveSession();
-    this.isJoined = false;
   }
 
   @action screenShare() {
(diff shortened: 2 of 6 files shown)


satyam73 (changes requested):
Also @SanketDhabarde for guest we will use HLS stream right?

vvaibhavdesai (approved):

app/constants/live.js
```diff
@@ -1,3 +1,7 @@
-export const ROLES = { guest: 'guest', host: 'host', presenter: 'presenter' };
+export const ROLES = {
+  guest: 'guest',
+  host: 'host',
+  presenter: 'presenter',
```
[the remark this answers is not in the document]
SanketDhabarde: I will remove it. Added during initial roles

app/controllers/live.js
```diff
@@ -28,7 +28,7 @@ export default class LiveController extends Controller {
     super(...arguments);
     setTimeout(() => {
```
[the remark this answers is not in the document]
SanketDhabarde: Sure

SanketDhabarde:
> Also @SanketDhabarde for guests we will use HLS stream right?

Yes, that will be when we allow multiple presenters and guests. Currently, the implementation is with one host and numerous guests.

app/services/live.js
```diff
@@ -48,9 +60,56 @@ export default class LiveService extends Service {
...
+      console.error(error);
+    }
+  }
+
+  async joinSession(userName, role, room) {
     try {
-      const token = await this.getToken(userName, userType);
+      const roomId =
+        ROLES.host === role ? await this.createRoom(userName) : room;
+      console.log({ roomId }); // For now use it to create link for guest
```
harshith-venkatesh, line 110: Why are we logging roomId here?
Is there a significance for the same

tests/acceptance/live-test.js
```diff
@@ -1,12 +1,12 @@
...
-import { visit, currentURL, waitFor, typeIn, click } from '@ember/test-helpers';
+import { visit, currentURL, typeIn, click } from '@ember/test-helpers';
 import { setupApplicationTest } from 'website-www/tests/helpers';
 
 module('Acceptance | live', function (hooks) {
   setupApplicationTest(hooks);
   // TODO - remove test for dev part when it goes to production
   test('skeleton and video should exists after join modal', async function (assert) {
-    assert.expect(11);
+    assert.expect(6);
```
harshith-venkatesh, line 9: Could you explain more about change in this test case

tests/acceptance/live-test.js
```diff
@@ -29,14 +29,5 @@ module('Acceptance | live', function (hooks) {
 
     await typeIn('[data-test-input-field]', 'sanket');
     await click('[data-test-button="live-join"]');
```
harshith-venkatesh, line 31: I see series of test cases removed after the clock of event, any particular reason for the same

config/environment.js
```diff
@@ -43,9 +40,6 @@ module.exports = function (environment) {
   if (environment === 'test') {
     // Testem prefers this...
     ENV.locationType = 'none';
```
harshith-venkatesh, line 42: The below credentials could be mishandled, ensure it doesn't get merged in general

app/services/live.js
SanketDhabarde: The host needs to share the invite link with the guests to join the live share. And we need the `roomId` to join. For now, we are getting `roomId` from the log. But I am fixing it in this issue #500 point 3.

config/environment.js
SanketDhabarde: I have removed this credential from this PR. It was needed before to make the 100ms APIs work. Now we are using our own backend API so, it is not needed. For security purposes, I have disabled the room ID.

tests/acceptance/live-test.js
SanketDhabarde: After the clock, the backend API call is going. Which makes the further assertion fails. We have added TODO to add the test cases for the Service.

tests/acceptance/live-test.js
SanketDhabarde: https://github.com/Real-Dev-Squad/website-www/pull/498#discussion_r1240623617

tests/acceptance/live-test.js
harshith-venkatesh: Should we create a ticket to address this?

harshith-venkatesh (approved):

tests/acceptance/live-test.js
SanketDhabarde: Yes sure I will create it. Thank you.

rohan09-raj (approved):

PR #527, opened 2023-07-15
[Ember-Migration] Added `numberOfHours` field in Join form

### LINKED ISSUE #516: Add number of Hours field

chandrareddy7:
### What is the change?
- Added InputBox component with label 'How many hours per week, are you willing to contribute?' and key 'numberOfHours'.
- Added validation code for numberOfHours.

### Is Development Tested?
- [x] Yes
- [ ] No

### DIFF
--- a/app/components/join-steps/step-three.js
+++ b/app/components/join-steps/step-three.js
@@ -42,7 +42,7 @@ export default class StepThreeComponent extends Component {
         validator(this.data.whyRds, 100) &&
         validator(this.data.foundFrom, 1) &&
         (e.target.name === 'numberOfHours'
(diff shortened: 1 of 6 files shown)


app/components/join-steps/step-three.js
rohan09-raj, line 1: I am able fill `200` value which is not in the range actually. Can you please check this ?
image

rohan09-raj: Preview button shouldn't enabled when value is not in range
image

chandrareddy7: Added validation in other commit @rohan09-raj

rohan09-raj (approved):

PR #551, opened 2023-08-04
fixed variable name in navbar.cc

fakhruddinkw:
due to variable name crashing in join.css and navbar.css the color of dropdown items on join.css is appearing black when they should be blue


Pratiyushkumar (approved):
LGTM

Issue #561, opened 2023-08-11
Ember tests failing issue, remove skip in tests


satyam73:
Ember tests failing issue, remove skip in tests which has been skipped in the PR - #560 

![image](https://github.com/Real-Dev-Squad/website-www/assets/85296770/613ebce5-8961-4808-a339-d6e65b1bcbea)
![image](https://github.com/Real-Dev-Squad/website-www/assets/85296770/3d2313af-643c-4956-ab1f-85143cee1730)

PR #564, opened 2023-08-11
Fix failing tests

prakashchoudhary07:
### Issue

#561 

stepper.js test was previously skipped, Track this in a separate issues

### Is Development Tested?

- [x] Yes
- [ ] No


app/constants/urls.js
```diff
@@ -1,24 +1,58 @@
...
 
-export const APPS = {
-  HOME: ENV.APPS?.HOME,
-  WELCOME: ENV.APPS?.WELCOME,
-  EVENTS: ENV.APPS?.EVENTS,
-  MEMBERS: ENV.APPS?.MEMBERS,
-  STATUS: ENV.APPS?.STATUS,
-  PROFILE: ENV.APPS?.PROFILE,
-  LIVE: 'live',
+const SCHEMA = 'https://';
```
satyam73, line 3: Isn't the name `SCHEMA` seems confusing?
Can we name it `PROTOCOL`?

satyam73 (changes requested):

app/constants/urls.js
prakashchoudhary07: https://developer.mozilla.org/en-US/docs/Learn/Common_questions/Web_mechanics/What_is_a_URL#scheme

prakashchoudhary07: I can change it its confusing

satyam73 (commented):
There is conflicts in the PR, please sync with the current branch.

app/constants/urls.js
satyam73: Hmm got it👍, we can keep it

satyam73 (dismissed):
LGTM

rohan09-raj (approved):

Issue #565, opened 2023-08-12
[Live-Site] - Sidebar peers map according to roles 


Pratiyushkumar:
Need to design the sidebar panel where peers who joined the stream, name shows up according to roles
![image](https://github.com/Real-Dev-Squad/website-www/assets/82165483/812e2088-3e1e-41a0-9171-ab6149bd7af1)

Pratiyushkumar:
PR merged https://github.com/Real-Dev-Squad/website-www/pull/583

PR #583, opened 2023-08-22
[Live-Site] - Sidebar peers map according to roles

Pratiyushkumar:
### Issue
- #565 

### What is the change?
The change is that we have separated participants on the basis of their roles who have joined the live streaming. If no user has joined the stream then it will so for each role as "No `rolename` in the stream". example - "No Mavens in the stream".

### Is Development Tested?

- [x] Yes
- [ ] No

### Test Coverage:
image

### Video Recording:

https://github.com/Real-Dev-Squad/website-www/assets/82165483/32ab0867-5a23-49a8-8985-7796b7de7a24


app/services/live.js
```diff
@@ -30,6 +30,9 @@ export default class LiveService extends Service {
   @globalRef('videoEl') videoEl;
   @tracked peers;
   @tracked isScreenShareOn;
+  @tracked isAnyMavenPresent = '';
+  @tracked isAnyModeratorPresent = '';
```
bhtibrewal, line 34: `isAnyMavenPresent` and `isAnyModeratorPresent`
variable starting with "is" mostly indicates that the variable holds a boolean value

yesyash (approved):

app/services/live.js
yesyash: - Agree with Bhavika here Should it be named to  `mavenRole`?

tests/integration/components/live-participants-test.js
```diff
@@ -0,0 +1,188 @@
...
+          roleName: 'maven',
+        },
+        {
+          id: 3,
+          name: 'Maven2',
+          roleName: 'maven',
+        },
+      ],
+      profilePic: 'profilepicurl',
+      isKickoutModalOpen: true,
```
bhtibrewal, line 106: please move this data to a separate file so it can be reused, it's being used repeatedly in this file also

app/services/live.js
```diff
@@ -30,6 +30,9 @@ export default class LiveService extends Service {
   @globalRef('videoEl') videoEl;
   @tracked peers;
   @tracked isScreenShareOn;
```
bhtibrewal, line 32: i think it should have a default value as false

app/services/live.js
```diff
@@ -297,6 +300,13 @@ export default class LiveService extends Service {
 
   async renderScreenVideoToPeers(peers) {
     this.peers = peers;
+    const maven = peers.find((peer) => peer.roleName === ROLES.maven);
+    const moderator = peers.find((peer) => peer.roleName === ROLES.moderator);
+    const guest = peers.find((peer) => peer.roleName === ROLES.guest);
```
bhtibrewal, line 305: we should optimize this, please think about how we can do this in one loop instead of 3 and make it scalable if  we add more roles

app/services/live.js
Pratiyushkumar: yes it should be but I have not written this code

app/services/live.js
Pratiyushkumar: i and Satyam tried doing but this was not working we had do like this

tests/integration/components/live-participants-test.js
Pratiyushkumar: done

app/services/live.js
Pratiyushkumar: if you have solution then please let me know

app/services/live.js
Pratiyushkumar: done

app/services/live.js
Pratiyushkumar: done

app/services/live.js
```diff
@@ -297,6 +300,20 @@ export default class LiveService extends Service {
 
   async renderScreenVideoToPeers(peers) {
     this.peers = peers;
+    const rolesToFind = [ROLES.maven, ROLES.moderator, ROLES.guest];
+    const foundPeers = {};
+
+    for (const roleName of rolesToFind) {
+      foundPeers[roleName] = peers.find((peer) => peer.roleName === roleName);
+    }
```
bhtibrewal, line 308: still we are running loop for each roleName, and shouldn't `foundPeers[roleName] ` be a list ?

app/services/live.js
```diff
@@ -297,6 +300,20 @@ export default class LiveService extends Service {
...
+
+    const maven = foundPeers[ROLES.maven];
+    const moderator = foundPeers[ROLES.moderator];
+    const guest = foundPeers[ROLES.guest];
+    this.mavenRole = maven?.roleName;
+    this.moderatorRole = moderator?.roleName;
+    this.guestRole = guest?.roleName;
```
bhtibrewal, line 315: if we are assigning object/array to these variable, then why are we keeping the default values as string

app/services/live.js
Pratiyushkumar: no, because we need to track the peer's array to if there is any role name with maven, moderator, or guest present in that, and if not then it should show a message.... I did try some options  but nothing worked

app/services/live.js
Pratiyushkumar: same as above

app/components/live-participants.js
```diff
@@ -0,0 +1,13 @@
+import Component from '@glimmer/component';
+import { ROLES } from '../constants/live';
+import { getOwner } from '@ember/application';
+import { inject as service } from '@ember/service';
+
+export default class LiveParticipantsComponent extends Component {
+  @service live;
```
satyam73, line 7: remove this @service as we're using getter for getting service below

app/components/live-sidebar.js
```diff
@@ -3,14 +3,21 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { ROLES } from '../constants/live';
 import { getOwner } from '@ember/application';
+import { inject as service } from '@ember/service';
 
 export default class LiveHeaderComponent extends Component {
+  @service live;
```
satyam73, line 9: same as above if live service is not used then remove this one too.

app/components/live-sidebar.js
```diff
@@ -3,14 +3,21 @@ import { action } from '@ember/object';
...
   @tracked isSideBarOpen = false;
   ROLES = ROLES;
-  get liveService() {
-    return getOwner(this).lookup('service:live');
+  constructor(...args) {
```
satyam73, line 12: We can also remove this constructor

app/services/live.js
satyam73: @bhtibrewal we are using @tracked variable for adding string which is roleName here

tests/constants/participants-data.js
satyam73, line 1: make sure exisiting tests doesn't fail if we've moved the constant in separate constant file

tests/integration/components/live-participants-test.js
```diff
@@ -0,0 +1,128 @@
...
+      @user='Mavens'
+      @role='maven'
+      @peers={{this.peers}}
+      @minimumParticipants=""
+      @openKickoutModal={{this.openKickoutModal}}
+    />`);
+    assert
+      .dom('[data-test-sidebar-body-role] .user')
+      .hasText('No Mavens in the stream');
+  });
```
satyam73, line 33: add spacing here

tests/integration/components/live-participants-test.js
```diff
@@ -0,0 +1,128 @@
...
+      @role='moderator'
+      @peers={{this.peers}}
+      @minimumParticipants=""
+      @openKickoutModal={{this.openKickoutModal}}
+    />`);
+    assert
+      .dom('[data-test-sidebar-body-role] .user')
+      .hasText('No Moderators in the stream');
+  });
+  test('renders No Guests In the stream Text', async function (assert) {
```
satyam73, line 54: add spacing here

tests/integration/components/live-participants-test.js
```diff
@@ -0,0 +1,128 @@
...
+      @role='guest'
+      @peers={{@peers}}
+      @minimumParticipants=""
+      @openKickoutModal={{this.openKickoutModal}}
+    />`);
+    assert
+      .dom('[data-test-sidebar-body-role] .user')
+      .hasText('No Guests in the stream');
+  });
+  test('renders Mavens Lists who joined Stream', async function (assert) {
```
satyam73, line 74: add spacing here

tests/integration/components/live-participants-test.js
```diff
@@ -0,0 +1,128 @@
...
+    await render(hbs`<LiveParticipants
+      @user='Mavens'
+      @role='maven'
+      @peers={{this.peers}}
+      @minimumParticipants="maven"
+      @openKickoutModal={{this.openKickoutModal}}
+    />`);
+    assert.dom('[data-test-sidebar-user="2"]').hasText('Maven1');
+  });
+  test('renders Moderators Lists who joined Stream', async function (assert) {
```
satyam73, line 92: add spacing here

tests/integration/components/live-participants-test.js
```diff
@@ -0,0 +1,128 @@
...
+    await render(hbs`<LiveParticipants
+      @user='Moderator'
+      @role='moderator'
+      @peers={{this.peers}}
+      @minimumParticipants="moderator"
+      @openKickoutModal={{this.openKickoutModal}}
+    />`);
+    assert.dom('[data-test-sidebar-user="3"]').hasText('Mod3');
+  });
+  test('renders Guests Lists who joined Stream', async function (assert) {
```
satyam73, line 110: add spacing here

tests/integration/components/live-participants-test.js
satyam73, line 1: there isn't test for @openKickoutModal function, try adding the test for the function too

satyam73 (changes requested):

tests/constants/participants-data.js
Pratiyushkumar: not failing

app/components/live-sidebar.js
Pratiyushkumar: fixed

app/components/live-sidebar.js
Pratiyushkumar: fixeed

app/components/live-participants.js
Pratiyushkumar: fixed

tests/integration/components/live-participants-test.js
Pratiyushkumar: done

tests/integration/components/live-participants-test.js
Pratiyushkumar: done

tests/integration/components/live-participants-test.js
Pratiyushkumar: done

tests/integration/components/live-participants-test.js
Pratiyushkumar: done

tests/integration/components/live-participants-test.js
Pratiyushkumar: done

tests/integration/components/live-participants-test.js
Pratiyushkumar: it is inside live.js controller I have just used it, we can take this later if required, not required in this PR

app/components/live-participants.hbs
```diff
@@ -0,0 +1,39 @@
+<div data-test-sidebar-body-role class='body__role-user'>
```
iamitprakash, line 1: not-eq function is being used to compare the @minimumParticipants variable and the @role variable. This is incorrect, because the not-eq function is used to compare two values and return true if they are not equal. In this case, the @minimumParticipants variable and the @role variable are both strings, so they cannot be compared using the not-eq function.

app/components/live-sidebar.hbs
```diff
@@ -40,36 +40,26 @@
...
-                    color='red'
-                  />
-                </button>
-              {{/if}}
-            </li>
-          {{/if}}
-        {{/each}}
-      </ul>
-    </div>
+    <LiveParticipants
```
iamitprakash, line 43: @minimumParticipants variable is being passed to the LiveParticipants component three times, but it is only being defined once. This means that the value of the @minimumParticipants variable will be the same for all three LiveParticipants components.

iamitprakash (changes requested):

iamitprakash (changes requested):
please add video from local system

PR #628, opened 2023-09-03
feat: add year in event card

### LINKED ISSUE #624: Event card on /event page
## Description: 

- On the events card page, I see that the date mentioned is `Sunday, April 20th, 12: 00pm IST` But thinking from a broader view, what if user comes here after two years? User will definitely be confused about in which `year` this event happened which is not a good user experience. I feel that we should add `year` also along with the existing date and time. 

## Potential Solution: 

- We can add the exact year along with other date and timeline on event card. 

## Screenshots: 

image

samyakshah3008:
### What is the change?

For a better user experience, we have added `year` as well along with the date of event. Earlier `year` was missing which can lead confusion to user considering the edge case where there are events with same date but with different years. 

### Is Development Tested?

- [X] Yes
- [ ] No

#### Before Screenshot: 

image

#### After Screenshot: 

image

### DIFF
--- a/events.html
+++ b/events.html
@@ -155,7 +155,7 @@
             Building backend services can be challenging, but at Real Dev Squad,
             our members tackle interesting problems regularly.
           </h3>
-          <h3 class="event-date-time">Sunday, April 25th, 12:00pm IST</h3>
+          <h3 class="event-date-time">Sunday, April 25th 2021, 12:00pm IST</h3>
           <h2 class="speaker-name">
             Swaraj Rajpure
             <span class="speaker-social-info linkedin-info">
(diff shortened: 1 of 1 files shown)


events.html
```diff
@@ -155,7 +155,7 @@ <h3 class="event-date-time">
             Building backend services can be challenging, but at Real Dev Squad,
             our members tackle interesting problems regularly.
           </h3>
-          <h3 class="event-date-time">Sunday, April 25th, 12:00pm IST</h3>
+          <h3 class="event-date-time">Sunday, April 25th 2021, 12:00pm IST</h3>
```
satyam73, line 158: How you got that this one event is from 2021?

satyam73 (changes requested):

events.html
samyakshah3008: I clicked on the register button and saw the event date and year

pallabez (approved):
LGTM

satyam73 (commented):
^

satyam73 (approved):
LGTM

rohan09-raj (approved):

Issue #623, opened 2023-09-01
Title upcoming events not up to the mark


samyakshah3008:
## Description: 

- On the events page of our community `website-www` in the `/events` page, there is a buggy behavior observed. We have listed a couple of events. 
1. Title is `upcoming events` but event has already happened. So can we come up with a different title of `past events`? We can also improve on the design with discussing from UI/UX and Design team. As of now title of "upcoming events" is not a good user experience. 
2. Card is disabled. Not sure why? 
3. Is the events we listed is up to date? Or more events happened previously? (If yes then we can create a separate issue ticket and work on it) 

## Potential Solution: 

- Let's change the title to previous events 
- Let's add a section of upcoming events and add a message that there are no upcoming events, please keep an eye on our social handles for all the latest updates. 

## Screenshots: 

![image](https://github.com/Real-Dev-Squad/website-www/assets/86982322/ca9e2bbd-3cf3-4856-be1b-6671243ac121)

samyakshah3008:
If the issue ticket looks good and we come to conclusion to fix it, then I would love to work on it.

samyakshah3008:
This issue ticket has been successfully completed via PR #629 hance closing the issue now.

PR #629, opened 2023-09-03
changed title of event page

samyakshah3008:
### Issue

Closes issue ticket #623 

### What is the change?

Previously the event page title was `upcoming events` but events which we are displaying are already happened. So changed the title to `Recap of past events` 

### Is Development Tested?

- [X] Yes
- [ ] No

### Before Change Screenshot: 

image

### After Change Screenshot: 

image

### DIFF
--- a/events.html
+++ b/events.html
@@ -1,209 +1,291 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
-<head>
-  <meta charset="UTF-8" />
-  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta property="og:title" content="Upcoming Events" />
+    <meta
+      property="og:description"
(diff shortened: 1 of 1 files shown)


events.html
```diff
@@ -145,7 +145,7 @@
         />
       </a>
       <section class="event-details">
-        <h2 class="event-upcoming">Upcoming Events</h2>
+        <h2 class="event-upcoming">Recap of our past events</h2>
```
satyam73, line 148: There are mulitple issues mentioned in the ticket, but there's very small change here.
Not getting this thing🤔

satyam73 (changes requested):

pallabez (approved):
LGTM

satyam73 (approved):
LGTM

events.html
satyam73: Got it @samyakshah3008

rohan09-raj, line 148: I think only `Past Events` will be better here, what do you think on this ?

samyakshah3008: Sure, changed.

rohan09-raj (approved):

Issue #617, opened 2023-08-31
Rendering for a host in sidebar as per other roles.


Pratiyushkumar:
Rendering for a host in the sidebar as per other roles.

PR #633, opened 2023-09-05
[Live-Site]- Rendering for a host in sidebar as per other roles.

Pratiyushkumar:
### Issue
- #617 

### Is Development Tested?

- [x] Yes
- [ ] No

### Test Coverage - 
image

### After Change Screenshots
image

### DIFF
--- a/tests/integration/components/live-participants-test.js
+++ b/tests/integration/components/live-participants-test.js
@@ -77,7 +77,7 @@ module('Integration | Component | live-participants', function (hooks) {
 
   test('renders host name In the stream ', async function (assert) {
     this.setProperties({
-      peers: mavenPeerData,
+      peers: hostPeer,
       profilePic: 'profilepicurl',
       isKickoutModalOpen: true,
     });
--- a/tests/integration/components/live-participants-test.js
+++ b/tests/integration/components/live-participants-test.js
@@ -11,6 +11,7 @@ import {
 
 module('Integration | Component | live-participants', function (hooks) {
   setupRenderingTest(hooks);
+
   test('renders No Maven In the stream Text', async function (assert) {
     this.setProperties({
       peers: hostPeer,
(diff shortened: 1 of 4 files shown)


tests/integration/components/live-participants-test.js
```diff
@@ -74,6 +75,25 @@ module('Integration | Component | live-participants', function (hooks) {
       .hasText('No Guests in the stream');
   });
 
+  test('renders host name In the stream ', async function (assert) {
+    this.setProperties({
+      peers: mavenPeerData,
```
satyam73, line 80: why using mavenPeerData here?

tests/integration/components/live-participants-test.js
```diff
@@ -74,6 +75,25 @@ module('Integration | Component | live-participants', function (hooks) {
...
+      this.isKickoutModalOpen = true;
+    });
+    await render(hbs`<LiveParticipants
+      @user='Hosts'
+      @role='host'
+      @peers={{this.peers}}
+      @minimumParticipants="host"
+      @openKickoutModal={{this.openKickoutModal}}
+    />`);
+    assert.dom('[data-test-sidebar-user="1"]').hasText('Ankush');
```
satyam73, line 94: More tests can be added

satyam73 (changes requested):

tests/integration/components/live-participants-test.js
Pratiyushkumar: please suggest if you think so

tests/integration/components/live-participants-test.js
Pratiyushkumar: fixed

tests/integration/components/live-participants-test.js
satyam73: Like currently you've only added tests for checking peers prop.
What about the rest of props?
What about the function you've passed?

Pratiyushkumar: I did not get your comment here, can you specify a little bit because I don't think any scenario is left here to test, but still let me know and if those are not that important then let'e merge this PR

prerana1821 (approved):
LGTM

satyam73 (approved):

Issue #615, opened 2023-08-31
Loading state while kicking out


Pratiyushkumar:
There should be a loader shown while kicking out the peers from a live stream, so when we kick peers, there is a delay between the request sent to the backend and the response from the backend, so to cover that delay we are putting a loader.

PR #641, opened 2023-09-13
[Live-Site] Feature/loading

Pratiyushkumar:
### Issue
- #615 

### What is the change?
There should be a loader shown while kicking out the peers from a live stream, so when we kick peers, there is a delay between the request sent to the backend and the response from the backend, so to cover that delay we are putting a loader.

### Is Development Tested?

- [x] Yes
- [ ] No

### Change video recording

https://github.com/Real-Dev-Squad/website-www/assets/82165483/fbf43083-4463-4c1d-9a14-16a8142f1bfa


app/components/live-participants.hbs
```diff
@@ -21,19 +21,33 @@
...
-                <IconifyIcon
-                  data-test-icon='remove'
-                  @icon='material-symbols:person-remove'
-                  width='36'
-                  height='36'
-                  color='red'
-                />
-              </button>
+              {{#if this.liveService.isUserRemoved}}
+                <div class='loading-container'>
```
prerana1821, line 25: Please follow the BEM convention for the class

app/styles/live-sidebar.module.css
```diff
@@ -141,6 +140,55 @@
...
 
+.loading-container {
+  display: flex;
+  flex-direction: row;
+}
+
+.loading span {
+  width: 8px;
+  height: 8px;
+  margin: 0 5px;
```
prerana1821, line 151: Please use rem for consistency.

prerana1821 (changes requested):
Hey @Pratiyushkumar,
image
I don't think we need both loading text and the animation. 
You can remove the "Loading" text. Also, can the space between the dots decrease a little?

app/components/live-participants.hbs
Pratiyushkumar: removed

app/styles/live-sidebar.module.css
Pratiyushkumar: done

Pratiyushkumar:
> Hey @Pratiyushkumar, image I don't think we need both loading text and the animation. You can remove the "Loading" text. Also, can the space between the dots decrease a little?

removed loading text only animation is there

satyam73:
The loader is overall goo can we decrease the size of bubbles, a little and make it pink color(for all the three bubbles).

satyam73 (dismissed):
LGTM

vinit717 (dismissed):

satyam73 (approved):

vinit717 (approved):

app/components/live-participants.hbs
```diff
@@ -24,20 +24,30 @@
...
-                    height='36'
-                    color='red'
-                  />
-                </button>
+                {{#if this.liveService.isUserRemoved}}
+                  <div class='loading'>
+                    <span></span>
+                    <span></span>
+                    <span></span>
+                  </div>
```
prakashchoudhary07, line 32: NIT: this can be broken into a separate component

Issue #865, opened 2024-02-06
Pre fill name for host and moderator(when logged in)


satyam73:
### Issue Description

We need to prefill the input of name if the user is logged in, in live site. It will make better UI/UX experience for the user since they don't have to enter name everytime when they're logged in.

### Expected Behavior

We need to prefill the input of name if the user is logged in, in live site. It will make better UI/UX experience for the user since they don't have to enter name everytime when they're logged in.

### Current Behavior

We have to enter name manually if we have to enter the live event.

### Screenshots

n/a

### Reproducibility

 - [x] This issue is reproducible
-  [ ] This issue is not reproducible

### Steps to Reproduce
- Go to /live with dev=true feature flag to see the issue

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [ ] Medium
 - [ ] Low

### Additional Information

[Include any additional context, logs, or details that might be helpful in addressing the issue.]

### Checklist

- [ ] I have read and followed the project's code of conduct.
- [ ] I have searched for similar issues before creating this one.
- [ ] I have provided all the necessary information to understand and reproduce the issue.
- [ ] I am willing to contribute to the resolution of this issue.

---

Thank you for contributing to our project! We appreciate your help in improving it.

jsuryakt:
@satyam73 Can you assign this to me

PR #873, opened 2024-02-15
Feat: prefill first name in input for events

jsuryakt:
Date: 16 Feb 2024

Developer Name: Jayasurya

---

## Issue Ticket Number:-

#865 

## Description:

Prefill first name of user in events input field

Is Under Feature Flag

- [x] Yes
- [ ] No

Database changes

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [x] No

### DIFF
--- a/tests/integration/components/live-join-test.js
+++ b/tests/integration/components/live-join-test.js
@@ -170,7 +170,13 @@ module('Integration | Component | live-join', function (hooks) {
     assert.dom('[data-test-live-join-title]').containsText('Create a Event');
   });
 
-  test('it should prefill name in input box', async function (assert) {
+  test('it should prefill first name in input box', async function (assert) {
+    const loginService = this.owner.lookup('service:login');
+    loginService.userData = { first_name: 'Jayasurya' };
+
+    const controller = this.owner.lookup('controller:live');
+    assert.deepEqual(controller.name, 'Jayasurya');
+
     const objToCheckFunctions = {
       isInputHandler: assert.ok(true, 'inputHandler is working fine!'),
       isClickHandlerWorks: assert.ok(true, 'clickHandler is working fine!'),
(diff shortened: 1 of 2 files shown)


iamitprakash (changes requested):
test case is missing

jsuryakt:
> test case is missing

There are test cases with name already set, do you mean e2e test?

tests/integration/components/live-join-test.js
```diff
@@ -169,4 +169,40 @@ module('Integration | Component | live-join', function (hooks) {
...
+  test('it should prefill name in input box', async function (assert) {
+    const objToCheckFunctions = {
+      isInputHandler: assert.ok(true, 'inputHandler is working fine!'),
+      isClickHandlerWorks: assert.ok(true, 'clickHandler is working fine!'),
+      isBackHandlerWorks: assert.ok(true, 'backHandler is working fine!'),
+    };
+    this.setProperties({
+      role: 'host',
+      name: 'Jayasurya',
+      roomCode: '',
```
ajoykumardas12, line 182: This test explicitly sets the name and then checks if name is rendered. Can we instead set login data and check if first name is rendered?

satyam73:
The input can be disabled then if name is already filled

jsuryakt:
> The input can be disabled then if name is already filled

Why? The user can edit it if needed and also if we disable the input, then there is no use of showing it at all, can just proceed with other fields.

tests/integration/components/live-join-test.js
ajoykumardas12: Controller name value assertion looks good to me.

ajoykumardas12:
> The input can be disabled then if name is already filled

If not logged in and joining as guest, name won't pre-filled. Not sure if it can/should be disabled when logged in.

ajoykumardas12 (approved):

iamitprakash (approved):

Issue #895, opened 2024-08-16
Archived users unable to use /join


ankushdharkar:
### Issue Description
When some new users (or even old users) who are not in our discord yet use the /join page, sometimes the cron marks them `archived` and stops them from filling the form

### Expected Behavior
Users who are not in discord are considered archived, but they should be able to fill the form

### Current Behavior
<img width="765" alt="image" src="https://github.com/user-attachments/assets/3a38227b-7957-4a2f-b42a-0d792bd07f0c">

### Reproducibility

 - [ ] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
1. Log in as an archived user
2. Try to fill the form on /join

### Severity/Priority
 - [x] Critical
 - [ ] High
 - [ ] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

PR #896, opened 2024-08-16
fix(895)!: temp fixes join page for archived users

ankushdharkar:
Date: Aug 16, 2024

Developer Name: Ankush Dharkar

---

## Issue Ticket Number:-

- #895

## Description:

Temporarily unblocks archived users to submit the /join form. Cron job seems to be marking users not in discord as archived and blocking them

Is Under Feature Flag

- [ ] Yes
- [x] No - This is a hotfix and needs to be available on prod

Database changes

- [ ] Yes
- [x] No

- [x] Yes - Unintended side-effects may occur from /join APIs now working for archived users. Risk: Minor (since backend should authorize correctly)
- [ ] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [x] No

### Add relevant Screenshot below ( e.g test coverage etc. )
Google Chrome 2024-08-16 16 32 26

### DIFF
--- a/app/components/stepper.js
+++ b/app/components/stepper.js
@@ -54,6 +54,8 @@ export default class StepperComponent extends Component {
       localStorage.setItem('first_name', this.login.userData.first_name);
       localStorage.setItem('last_name', this.login.userData.last_name);
       this.incrementStep();
+    } else {
+      alert('You must be logged in to continue');
     }
   }
 
--- a/app/templates/join.hbs
+++ b/app/templates/join.hbs
@@ -5,7 +5,8 @@
       <Fa-Icon @size='2x' @icon='circle-notch' @spin={{true}} />
     </div>
   {{else}}
-    {{#if this.login.userData.roles.archived}}
+    {{!-- Quick hack for unblocking everyone - https://github.com/Real-Dev-Squad/website-www/issues/895 --}}
+    {{#if false}} {{!-- {{#if this.login.userData.roles.archived}} --}}
       <OnboardingCard>
         <h3 data-test-archived-heading class='archived__heading'>Archived User</h3>
         <p data-test-archived-message class='archived__message'>


app/components/stepper.js
```diff
@@ -54,6 +54,8 @@ export default class StepperComponent extends Component {
       localStorage.setItem('first_name', this.login.userData.first_name);
       localStorage.setItem('last_name', this.login.userData.last_name);
       this.incrementStep();
+    } else {
+      alert('You must be logged in to continue');
```
ankushdharkar, line 58: Another small UX improvement. The button click was silently failing when users were not logged in

prakashchoudhary07:
Will have to check, If we need to do backend changes to accept response from archived users

ankushdharkar:
@prakashchoudhary07 Already tested the flow by marking myself as `archived`. Was able to submit the form through and through. Let's merge this and get a real user to test this for us

app/components/stepper.js
entropyReverser: Nice. I really thought the page was just broken.

prakashchoudhary07 (approved):

app/components/stepper.js
entropyReverser: Wait, just read the description. I thought this was a fix that hitting the "Start" button does nothing if you are not logged in, and was fixing that.

entropyReverser: Video: https://youtu.be/d5f0NCNOLvg

ankushdharkar: @entropyReverser That doesn't require a fix as the workaround is to log in using the "Sign in with GitHub" button on top-right
CleanShot 2024-08-16 at 17 06 38@2x

entropyReverser: Yeah. I know, but I would have liked an alert like this of  "Please log in to continue"

ankushdharkar: @entropyReverser That's exactly what this UX adds :)

Issue #970, opened 2025-01-05
Update Homepage - Jan 2025


ankushdharkar:
### Issue Description
- Add/update details on how to join section for 2025
- Add all the mentors on the homepage - Amit, Prakash, Yash

### Severity/Priority
 - [x] Critical

### Checklist
- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x]  I am willing to contribute to the resolution of this issue.
- [x] I have provided all the necessary information to understand and reproduce the issue.

ankushdharkar:
@prakashchoudhary07 This ticket is not yet ready for closing. Reopening it.

PR #994, opened 2025-02-02
feat(970): adds new 'how to join' message

ankushdharkar:
Date:  Feb 2, 2025

Developer Name: Ankush Dharkar

---

## Issue Ticket Number:-

- #970 

## Description:

Adds the new message regarding how to join the RDS server.

### Main page:
CleanShot 2025-02-02 at 17 07 31@2x

### Subscribe page

With FF ON:
image

Without FF:
image

Is Under Feature Flag

- [x] Yes
- [ ] No

Database changes

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [x] No


app/components/join-section.hbs
```diff
@@ -41,13 +41,12 @@
     {{!-- NOTE: This has the new code on feature flag off --}}
     {{!-- If something goes wrong, we can enable a flag to restore previous behaviour --}}
     {{!-- TODO: Clean up by Jan, 2025 --}}
+    {{#if this.isDevMode}}
     <p data-test-join-later-text class='join__later-text'>
-      {{#if this.isDevMode}}
       We're currently not accepting new applications. Please come back later.
```
ankushdharkar, line 46: Since the join section is styled separately from [subscribe page](https://github.com/Real-Dev-Squad/website-www/pull/994/files#diff-05efdc7e5350b3b112828cb661de492625904672c634d85e702321f8f4c63a38R43), have to pass the individual classes to the common notice component for applying the specific styles

app/templates/subscribe.hbs
```diff
@@ -36,16 +36,16 @@
       {{else}}
         <h1>Join the Real Dev Squad Community</h1>
 
+        {{!-- NOTE: This has the new code on feature flag off. --}}
+        {{!-- If something goes wrong, we can enable a flag to restore previous behaviour --}}
+        {{!-- TODO: Clean up by Jan, 2025 --}}
```
AnujChhikara, line 41: I noticed the TODO from the last PR about cleaning this up by Jan 2025. Since that deadline has passed and the cleanup is still pending, should we update this TODO?

tests/integration/components/notice-for-joining-test.js
AnujChhikara, line 1: I see that the tests have been updated, which makes sense. Since the dev flag still allows viewing the old behavior, do you think it would be useful to add a test to verify that the old text is rendered when dev=true , or would that be overkill?

ankushdharkar: Good point. We should actually test both FF on and off. BUT, in this case, if you switch on the FF, the whole www page changes to a new design, which doesn't have the messaging regarding joining the squad

app/templates/subscribe.hbs
ankushdharkar: Will update the date. New changes are going to what the FF is guarding, so advisable to keep the FF for some more time

tejaskh3:
Please update the branch,  this is our of date

tejaskh3 (changes requested):
can we reduce the gap between sentences in the referral message to improve readability and visual consistency (on the main page) ?

ankushdharkar:
@tejaskh3 The spacing is fine for this change. In future, the whole page will be redesigned anyway (switch on FF to see the new page)

tejaskh3 (approved):

tejaskh3:
> @tejaskh3 The spacing is fine for this change. In future, the whole page will be redesigned anyway (switch on FF to see the new page)

fine

app/components/join-section.hbs
```diff
@@ -40,14 +40,13 @@
...
-      {{else}}
-      <NoticeForJoining />
-      {{/if}}
     </p>
-
+    {{else}}
+    <NoticeForJoining data-test-join-later-text class="join__later-text"/>
```
MehulKChaudhari, line 49: I have noticed that the `<p>` and the component `NoticeForJoining` has same data test id's. 

we can keep the ID's different to avoid ambiguity and that will make it easier to target the correct element during testing

ankushdharkar: I would like to differ. The final element that has to be tested remains the same, whether FF is on or off :)

tejaskh3 (approved):
LGTM!

But before merging please create ticket for the cleanups you will be going to do.

AnujChhikara (approved):
LGTM

Achintya-Chatterjee (approved):

Issue #1088, opened 2025-11-05
Feature: Design UI for Application Tracking Page


AnujChhikara:
### Overview

- Create the UI for the Applications and Application Details pages for onboarding applications. This task is UI only and will use mock data - no backend integration yet.

### Requirements

- Applications Page

Show a list of applications (use mock data)

Click on an application → navigate to Application Details page

- Application Details Page

 Show the details overview of the applications with admin action buttons

PR #1090, opened 2025-11-06
feat: setup the basic route and navigation for the applications and application detail page

AnujChhikara:
Date: 7 Nov 2025

Developer Name: @AnujChhikara @MayankBansal12 

---

## Issue Ticket Number:-
- #1088

## Description:

- Set up the basic route and navigation for the applications and application detail page
- Test PR- #1100 

Is Under Feature Flag

- [x] Yes
- [ ] No

Database changes

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [x] No

### Add relevant Screenshot below ( e.g test coverage etc. )

https://github.com/user-attachments/assets/e16bfda1-d44e-4df9-9487-dc6ecbdc0acc

### DIFF
--- a/app/components/header.hbs
+++ b/app/components/header.hbs
@@ -64,6 +64,7 @@
       {{#if (and @dev this.isSuperUser)}}
         <li>
           <LinkTo
+            data-test-applications
             @route="applications"
             @query={{hash dev="true"}}
             class="nav__element"
--- a/tests/integration/components/header-test.js
+++ b/tests/integration/components/header-test.js
@@ -223,4 +223,65 @@ module('Integration | Component | header', function (hooks) {
     assert.dom('[data-test-dropdown-toggle]').doesNotExist();
     assert.dom('[data-test-login]').doesNotExist();
   });
+
+  test('applications link is visible for super user under dev flag', async function (assert) {
+    this.owner.register(
+      'service:login',
+      {
+        userData: { roles: { super_user: true } },
+      },
+      { instantiate: false },
+    );
+
+    this.setProperties({
+      firstName: 'John',
+      isLoggedIn: true,
+      isLoading: false,
+      dev: true,
+    });
+
+    this.set('signOut', () => {
+      this.isLoggedIn = false;
+    });
+
+    await render(hbs`
+      <Header
+        @firstName={{this.firstName}}
+        @isLoggedIn={{this.isLoggedIn}}
+        @isLoading={{this.isLoading}}
+        @signOut={{this.signOut}}
+        @dev={{this.dev}}
+      />
+    `);
+
+    assert.dom('[data-test-applications]').exists();
+    assert
+      .dom('[data-test-applications]')
+      .hasAttribute('href', '/applications?dev=true');
+  });
+
+  test('applications link is not visible for normal user under dev flag', async function (assert) {
+    this.setProperties({
+      firstName: 'John',
+      isLoggedIn: true,
+      isLoading: false,
+      dev: true,
+    });
+
+    this.set('signOut', () => {
+      this.isLoggedIn = false;
+    });
+
+    await render(hbs`
+      <Header
+        @firstName={{this.firstName}}
+        @isLoggedIn={{this.isLoggedIn}}
+        @isLoading={{this.isLoading}}
+        @signOut={{this.signOut}}
+        @dev={{this.dev}}
+      />
+    `);
+
+    assert.dom('[data-test-applications]').doesNotExist();
+  });
 });
--- a/tests/unit/controllers/applications-test.js
+++ b/tests/unit/controllers/applications-test.js
@@ -1,0 +1,12 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'website-www/tests/helpers';
+
+module('Unit | Controller | applications', function (hooks) {
+  setupTest(hooks);
+
+  test('it exists', function (assert) {
+    assert.expect(1);
+    const controller = this.owner.lookup('controller:applications');
+    assert.ok(controller, 'The applications controller exists');
+  });
+});
--- a/tests/unit/routes/applications-test.js
+++ b/tests/unit/routes/applications-test.js
@@ -1,0 +1,125 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'website-www/tests/helpers';
+import sinon from 'sinon';
+import {
+  APPLICATIONS_URL,
+  SELF_USER_PROFILE_URL,
+} from 'website-www/constants/apis';
+
+const APPLICATIONS_SIZE = 6;
+
+module('Unit | Route | applications', function (hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.fetchStub = sinon.stub(window, 'fetch');
+    this.route = this.owner.lookup('route:applications');
+    sinon.stub(this.route.router, 'replaceWith');
+    sinon.stub(this.route.toast, 'error');
+  });
+
+  hooks.afterEach(function () {
+    this.fetchStub.restore();
+    sinon.restore();
+  });
+
+  test('it exists', function (assert) {
+    assert.expect(1);
+    const route = this.owner.lookup('route:applications');
+    assert.ok(route, 'The applications route exists');
+  });
+
+  test('redirects to page-not-found when dev flag is not true', function (assert) {
+    const transition = { to: { queryParams: { dev: 'false' } } };
+
+    this.route.beforeModel(transition);
+
+    assert.ok(
+      this.route.router.replaceWith.calledOnceWith('/page-not-found'),
+      'Redirected to /page-not-found when dev is not true',
+    );
+  });
+
+  test('allows access when dev flag is true', function (assert) {
+    const transition = { to: { queryParams: { dev: 'true' } } };
+
+    this.route.beforeModel(transition);
+
+    assert.ok(
+      this.route.router.replaceWith.notCalled,
+      'No redirection occurs when dev query param is true',
+    );
+  });
+
+  test('returns null when transitioning to applications.detail', async function (assert) {
+    const transition = { to: { name: 'applications.detail' } };
+
+    const result = await this.route.model({}, transition);
+
+    assert.strictEqual(result, null, 'Returns null for detail route');
+    assert.ok(this.fetchStub.notCalled, 'No API calls made for detail route');
+  });
+
+  test('fetches applications successfully', async function (assert) {
+    const mockApplications = [
+      { id: 1, userId: 'user1' },
+      { id: 2, userId: 'user2' },
+    ];
+
+    this.fetchStub
+      .onCall(0)
+      .resolves(
+        new Response(JSON.stringify({ first_name: 'John' }), { status: 200 }),
+      );
+
+    this.fetchStub.onCall(1).resolves(
+      new Response(JSON.stringify({ applications: mockApplications }), {
+        status: 200,
+      }),
+    );
+
+    const transition = { to: { name: 'applications' } };
+    const result = await this.route.model({}, transition);
+
+    assert.deepEqual(result, mockApplications, 'Returns applications from API');
+    assert.ok(
+      this.fetchStub.firstCall.calledWith(
+        SELF_USER_PROFILE_URL,
+        sinon.match.object,
+      ),
+      'First API call is made to check user profile',
+    );
+    assert.ok(
+      this.fetchStub.secondCall.calledWith(
+        APPLICATIONS_URL(APPLICATIONS_SIZE),
+        sinon.match.object,
+      ),
+      'Second API call is made to fetch applications',
+    );
+  });
+
+  test('displays error toast and redirects on 401 response', async function (assert) {
+    this.fetchStub.resolves(new Response(JSON.stringify({}), { status: 401 }));
+
+    const transition = { to: { name: 'applications' } };
+    const result = await this.route.model({}, transition);
+
+    assert.strictEqual(result, null, 'Returns null for 401');
+    assert.ok(this.route.toast.error.calledOnce, 'Error toast is displayed');
+  });
+
+  test('displays error toast on API error', async function (assert) {
+    this.fetchStub
+      .onCall(0)
+      .resolves(new Response(JSON.stringify({}), { status: 200 }));
+    this.fetchStub
+      .onCall(1)
+      .resolves(new Response(JSON.stringify({}), { status: 500 }));
+
+    const transition = { to: { name: 'applications' } };
+    const result = await this.route.model({}, transition);
+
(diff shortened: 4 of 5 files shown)


app/templates/admin/applications.hbs
```diff
@@ -0,0 +1,312 @@
...
+        >
+          Next
+        </button>
+      </div>
+    {{/if}}
+  {{else}}
+    <div data-test-unauthorized class="admin-applications__unauthorized">
+      <p>You are not authorized to view this page.</p>
+    </div>
+  {{/if}}
```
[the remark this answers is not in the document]
AnujChhikara: - will add in next pr as this pr will get hard to review that why split it into multiple prs

app/styles/app.css
```diff
@@ -57,6 +57,8 @@
 @import url("qrcode.module.css");
 @import url("confirm-modal.module.css");
 @import url("new-signup.module.css");
+@import url("admin-applications.module.css");
+@import url("application-card.module.css");
```
Hariom01010: These files are being imported and not defined in ths PR. Won't it break the application or will the CSS PR will be merged before this PR

app/templates/admin/applications.hbs
```diff
@@ -0,0 +1,312 @@
...
+        data-test-summary-card-total
+        class="admin-applications__summary-card admin-applications__summary-card--total"
+      >
+        <div class="admin-applications__summary-card-icon">
+          <Fa-Icon @icon="file-alt" @size="2x" />
+        </div>
+        <div class="admin-applications__summary-card-content">
+          <p class="admin-applications__summary-card-label">Total Applications</p>
+          <h2 class="admin-applications__summary-card-value">
+            {{this.totalApplicationsCount}}
```
Hariom01010: Where are these template variables (e.g., isLoading, isSuperUser, totalApplicationsCount) coming from?

app/constants/apis.js
```diff
@@ -42,3 +42,15 @@ export const SELF_USER_PROFILE_URL = `${APPS.API_BACKEND}/users?profile=true`;
 export const QR_AUTHORIZATION_STATUS_URL = `${APPS.API_BACKEND}/auth/qr-code-auth/authorization_status`;
 
 export const USER_AUTHENTICATED_DEVICES_URL = `${APPS.API_BACKEND}/auth/device`;
+
+export const APPLICATIONS_LIST_URL = (size = 10, page = 1) => {
```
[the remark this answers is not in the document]
AnujChhikara: - they are done based on the currently api structure will be changed or verify when integrating the actual api

Hariom01010 (approved):

app/routes/admin/applications.js
```diff
@@ -0,0 +1,63 @@
...
+export default class AdminApplicationsRoute extends Route {
+  @service login;
+  @service router;
+  @service toast;
+  @service fastboot;
+
+  beforeModel() {
+    if (this.fastboot.isFastBoot) {
+      return;
+    }
```
MayankBansal12: do we really need this check for fastboot? can you please confirm what the behaviour was without this cond?

app/routes/admin/applications.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!this.login.isLoggedIn) {
+      this.router.transitionTo('/login');
+      return;
+    }
+
+    if (this.login.userData && !this.login.userData?.roles?.super_user) {
+      this.router.transitionTo('/page-not-found');
+    }
+  }
```
MayankBansal12: instead of role guarding for every route, we can put this check in parent /admin

eg: 

```js
admin.hbs

{{#if this.unauthorized}}
  <UnauthorizedMessage @message="You are not authorized to access admin pages." />
{{else}}
  {{outlet}}
{{/if}}
```

```js 
admin.js

beforeModel() {
  if (!this.login.isLoggedIn || !this.login.userData?.roles?.admin) {
    this.set('unauthorized', true);
  }
}

```

app/constants/apis.js
```diff
@@ -42,3 +42,15 @@ export const SELF_USER_PROFILE_URL = `${APPS.API_BACKEND}/users?profile=true`;
...
+  return `${APPS.API_BACKEND}/applications?size=${size}&page=${page}`;
+};
+
+export const APPLICATION_DETAIL_URL = (id) => {
+  return `${APPS.API_BACKEND}/applications/${id}`;
+};
```
MayankBansal12: please use dev flag here, since initially all these APIs will use feature flag

app/routes/admin/applications.js
```diff
@@ -0,0 +1,63 @@
...
+
+  async model(params) {
+    if (this.fastboot.isFastBoot) {
+      return { applications: [], total: 0, page: 1, size: 10 };
+    }
```
MayankBansal12: - not sure we require this check 
- also, we are already checking for isFastBoot in beforeModel, do we need this here too?

app/routes/admin/applications.js
```diff
@@ -0,0 +1,63 @@
...
+
+    try {
+      const size = params.size || 10;
+      const page = params.page || 1;
```
MayankBansal12:
```suggestion
      const { page = 1, size = 10 } = params;
```

app/routes/admin/applications.js
```diff
@@ -0,0 +1,63 @@
...
+      return {
+        applications: data.applications || [],
+        total: data.total || 0,
+        page: data.page || 1,
+        size: data.size || 10,
+      };
+    } catch (error) {
+      console.error('Error fetching applications:', error);
+      this.toast.error('Failed to load applications', 'Error', TOAST_OPTIONS);
+      return { applications: [], total: 0, page: 1, size: 10 };
```
MayankBansal12: do we need to return default response here? not throwing an error like in line 47?

Also define a constant for default response if we are going to use it more than once

app/templates/admin/applications.hbs
```diff
@@ -0,0 +1,308 @@
...
+        <button
+          data-test-next-page
+          type="button"
+          class="admin-applications__pagination-button"
+          disabled={{this.isNextDisabled}}
+          {{on "click" this.goToNextPage}}
+        >
+          Next
+        </button>
+      </div>
```
MayankBansal12: why are we not going for infinite scrolling for viewing all applications?

app/constants/apis.js
AnujChhikara: - why, do we want to keep the existing functionality of the api same without dev flag ?

MayankBansal12: Applications with the new format will be under the dev flag at first. We can make this change in follow-up PR for integration, but i would suggest using dev flag in this PR itself.

app/controllers/applications/detail.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+
+      if (!response.ok) {
+        const errorMessage = `Failed to fetch application: ${response.status}`;
+        console.error(errorMessage);
+        this.toast.error(errorMessage, 'Error!', TOAST_OPTIONS);
+        this.error = errorMessage;
+        this.isLoading = false;
+        return;
+      }
```
MayankBansal12: can we convert this into a single if cond?

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+import redirectAuth from '../../utils/redirect-auth';
+import { ERROR_MESSAGES } from '../../constants/error-messages';
+
+export default class ApplicationsDetailRoute extends Route {
+  @service toast;
+  @service fastboot;
+  @service router;
+
+  beforeModel(transition) {
+    if (transition?.to?.queryParams?.dev !== 'true') {
```
MayankBansal12: i guess there's already a variable defined for dev in services, see if we can reuse here

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+      const userResponse = await fetch(
+        `${APPS.API_BACKEND}/users?profile=true`,
+        {
+          credentials: 'include',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+          },
+        },
+      );
```
MayankBansal12: same as above, reuse util func here

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+
+      if (userResponse.status === 401) {
+        this.toast.error(ERROR_MESSAGES.notLoggedIn, '', TOAST_OPTIONS);
+        setTimeout(redirectAuth, 2000);
+        return { redirect: true, reason: 'unauthorized' };
+      }
+
+      if (!userResponse.ok) {
+        throw new Error(`HTTP error! status: ${userResponse.status}`);
+      }
```
MayankBansal12: please add null check for `userResponse`

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+        return { redirect: true, reason: 'forbidden' };
+      }
+
+      const response = await fetch(APPLICATION_BY_ID_URL(params.id), {
+        credentials: 'include',
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/json',
+        },
+      });
```
MayankBansal12: same here

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+
+  afterModel(model) {
+    if (model?.redirect) {
+      if (model.reason === 'not_found') {
+        this.router.replaceWith('/page-not-found');
+      } else {
+        this.router.replaceWith('/page-not-found');
+      }
+    }
+  }
```
MayankBansal12: this can be avoided if we simply redirect from above instead of passing here, do consider it!

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+
+  setupController(controller, model) {
+    super.setupController(controller, model);
+
+    if (!model?.redirect) {
+      controller.isLoading = false;
+      controller.applicationData = model.applicationData;
+      controller.error = model.error;
+    }
+  }
```
MayankBansal12: why is this required here?

app/routes/applications/detail.js
```diff
@@ -0,0 +1,131 @@
...
+
+  async model(params) {
+    if (this.fastboot.isFastBoot) {
+      return { id: params.id };
+    }
+
+    try {
+      const userResponse = await fetch(
+        `${APPS.API_BACKEND}/users?profile=true`,
```
MayankBansal12: there's a constant already defined that can be reused here

app/routes/applications.js
```diff
@@ -0,0 +1,111 @@
...
+    if (this._isLoading) {
+      return;
+    }
+
+    this._isLoading = true;
+    controller.isLoading = true;
+
+    try {
+      const userResponse = await fetch(
+        `${APPS.API_BACKEND}/users?profile=true`,
```
MayankBansal12: use constant here too

app/routes/applications/detail.js
```diff
@@ -0,0 +1,59 @@
...
+      const applicationResponse = await fetch(APPLICATION_BY_ID_URL(params.id));
+
+      if (!applicationResponse.ok) {
+        if (applicationResponse.status === 404) {
+          this.toast.error('Application not found', 'Error!', TOAST_OPTIONS);
+          this.router.replaceWith('/page-not-found');
+          return null;
+        }
+        throw new Error(`HTTP error! status: ${applicationResponse.status}`);
+      }
```
MayankBansal12: if it's going to be nested then they can be seperate if conditions

app/routes/applications.js
```diff
@@ -0,0 +1,67 @@
...
+    try {
+      const userResponse = await fetch(SELF_USER_PROFILE_URL, {
+        credentials: 'include',
+      });
+
+      if (userResponse.status === 401) {
+        this.toast.error(ERROR_MESSAGES.notLoggedIn, '', TOAST_OPTIONS);
+        setTimeout(redirectAuth, 2000);
+        this.router.replaceWith('/page-not-found');
+        return [];
```
MayankBansal12: we were returning null above and empty array, let's be consistent here

app/routes/applications.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      const userData = await userResponse.json();
+
+      if (!userData?.roles?.super_user) {
+        this.router.replaceWith('/page-not-found');
+        return null;
+      }
+
+      const applicationsResponse = await fetch(APPLICATIONS_URL(6));
```
MayankBansal12: what is 6 here? and why is it hardcoded?

app/controllers/applications/detail.js
```diff
@@ -0,0 +1,6 @@
+import Controller from '@ember/controller';
+import { service } from '@ember/service';
+
+export default class ApplicationsDetailController extends Controller {
+  @service router;
```
MayankBansal12: question: why we do need to use router here? 
is it not working without it?

app/controllers/applications.js
MayankBansal12: do we need this file now? or leaving it here for following PRs?

app/routes/applications.js
```diff
@@ -0,0 +1,72 @@
...
+  beforeModel(transition) {
+    if (transition?.to?.queryParams?.dev !== 'true') {
+      this.router.replaceWith('/page-not-found');
+    }
+  }
+
+  async model(params, transition) {
+    if (transition?.to?.name === 'applications.detail') {
+      return null;
+    }
```
MayankBansal12: this doesn't seem right, why do we need return here?

generally you don't need to return like this for nested routes, what's the behaviour without this?

also doesn't ember's automatically handle this, do we really need to define this explicitly?

app/routes/applications.js
```diff
@@ -0,0 +1,72 @@
...
+        APPLICATIONS_URL(APPLICATIONS_SIZE),
+      );
+
+      if (!applicationsResponse.ok) {
+        throw new Error(`HTTP error! status: ${applicationsResponse.status}`);
+      }
+
+      const applicationsData = await applicationsResponse.json();
+      const applications = applicationsData.applications || [];
+      return applications;
```
MayankBansal12:
```suggestion
      return applicationsData.applications || [];
```

app/routes/applications.js
```diff
@@ -0,0 +1,61 @@
...
+
+const APPLICATIONS_SIZE = 6;
+
+export default class ApplicationsRoute extends Route {
+  @service toast;
+  @service router;
+
+  queryParams = {
+    dev: { refreshModel: true },
+  };
```
Suvidh-kaushik: what is this?

Suvidh-kaushik (approved):

iamitprakash (approved):

